PR #3, opened 2019-02-20
Add zh-tw to lib/i18n

johnnychhsu:
init commit

### DIFF
--- a/lib/i18n/traditional-chinese.js
+++ b/lib/i18n/traditional-chinese.js
@@ -9,29 +9,6 @@
 // See COPYING for details
 "use strict";
 
-function replaceMeMy(sentence) {
-    sentence = sentence.replace(/\b((?!(?:通知|告知)\b)[a-zA-Z0-9]+) me\b/g, '$1 他們');
-
-    return sentence.replace(/\b(我|我的)\b/g, (what) => {
-        switch(what) {
-        case '我':
-            return '他';
-        case '我的':
-            return '他們的';
-        default:
-            return what;
-        }
-    });
-}
-
-function postprocessSynthetic(sentence, program) {
-    if (program.isProgram && program.principal !== null)
-        sentence = replaceMeMy(sentence);
-
-    return sentence.replace(/ (新的|新) (他們的|我的|一個) /, (_new, what) => ` ${what} ${_new} `)
-        .trim();
-}
-
 const ARGUMENT_NAME_OVERRIDES = {};
 
 const IGNORABLE_TOKENS = {
@@ -77,7 +54,6 @@ function isValidParaphrasePair(word, paraphrase) {
 }
 
 module.exports = {
-    postprocessSynthetic,
     detokenize,
 
     ARGUMENT_NAME_OVERRIDES,
--- a/lib/i18n/traditional-chinese.js
+++ b/lib/i18n/traditional-chinese.js
@@ -55,23 +55,7 @@ for (let abbr of ABBREVIATIONS) {
 
 }
 
-const SPECIAL_TOKENS = {
-    '.': '.',
-    ',': ',',
-    'n\'t': 'n\'t',
-    '\'s': '\'s',
-    '?': '?',
-
-    // right/left round/curly/square bracket
-    '-rrb-': ')',
-    '-lrb-': ' (',
-    '-rcb-': '}',
-    '-lcb-': ' {',
-    '-rsb-': ']',
-    '-lsb-': ' [',
-};
-
-function detokenize(sentence, prevtoken, token) {
+function detokenize(buffer, prevtoken, token) {
     // join without space
     return buffer + token;
 }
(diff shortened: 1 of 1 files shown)


lib/i18n/traditional-chinese.js
```diff
@@ -0,0 +1,107 @@
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of Genie
+//
+// Copyright 2018-2019 The Board of Trustees of the Leland Stanford Junior University
```
gcampax: You don't work for Stanford, so the copyright should be NTU?

lib/i18n/traditional-chinese.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+    });
+}
+
+function postprocessSynthetic(sentence, program) {
+    if (program.isProgram && program.principal !== null)
+        sentence = replaceMeMy(sentence);
+
+    return sentence.replace(/ (新的|新) (他們的|我的|一個) /, (_new, what) => ` ${what} ${_new} `)
+        .trim();
```
gcampax: This postprocessing step is for artifacts in the English templates, I don't know if you'll need it for Chinese - you won't know until you write the templates, really.

lib/i18n/traditional-chinese.js
```diff
@@ -0,0 +1,107 @@
...
+//
+// Copyright 2018-2019 The Board of Trustees of the Leland Stanford Junior University
+//
+// Author: Giovanni Campagna <gcampagn@cs.stanford.edu>, Johnny Hsu <johnny.chhsu01@gmail.com>
+//
+// See COPYING for details
+"use strict";
+
+function replaceMeMy(sentence) {
+    sentence = sentence.replace(/\b((?!(?:通知|告知)\b)[a-zA-Z0-9]+) me\b/g, '$1 他們');
```
gcampax: I tried to dig into git history to figure out what this line is doing but I can't quite understand it. I wonder if we can drop it.
In any case, `[a-zA-Z0-9]+` is a character class that matches a single word in English, but it won't do in Chinese, where characters are not a-z. If you want to replicate it exactly, you'll need to use the CJK ideograph range, possibly as a Unicode character class (and adding the `u` flag to the regexp).
Also, you'll probably want 我 instead of `me`?

Also, here and later, English uses `their` as gender neutral *singular* pronoun (because it's not easy to find out the gender of the replaced username at this stage), but won't do for Chinese. Maybe just use 他 and forget about it?

lib/i18n/traditional-chinese.js
```diff
@@ -0,0 +1,107 @@
...
+
+    return sentence.replace(/\b(我|我的)\b/g, (what) => {
+        switch(what) {
+        case '我':
+            return '他';
```
gcampax: The fact that Chinese does not distinguish `me` vs `i` might be a problem here...

lib/i18n/traditional-chinese.js
```diff
@@ -0,0 +1,107 @@
...
+    '?': '?',
+
+    // right/left round/curly/square bracket
+    '-rrb-': ')',
+    '-lrb-': ' (',
+    '-rcb-': '}',
+    '-lcb-': ' {',
+    '-rsb-': ']',
+    '-lsb-': ' [',
+};
```
gcampax: You don't need this (and indeed, eslint complains that you have it)

gcampax:
Also, don't back merge (from master into the branch). Instead, develop into your branch. If you need something from master, or if the branch does not merge cleanly, do a rebase and force-push into the branch, then merge the branch into master after the tests pass.

lib/i18n/traditional-chinese.js
johnnychhsu: That's right. Thank you!

johnnychhsu:
Sure! rebased the branch.

lib/i18n/traditional-chinese.js
johnnychhsu: We plan to use a blood pressure measuring machine as our demo device for Chinese, I think we can come up with some template with it first then see what will we need.

lib/i18n/traditional-chinese.js
johnnychhsu: Thank you!

lib/i18n/traditional-chinese.js
johnnychhsu: Maybe we can drop this here temporary, try some templates and see what we need?
In Chinese, there are 他 and 她, but yeah it's not a big deal in Chinese, we can use 他.

lib/i18n/traditional-chinese.js
johnnychhsu: Why ? I think we can use 我 and 他 to cover all cases if we need the postprocess here ?

PR #4, opened 2019-02-20
constant.genie for zh-tw

johnnychhsu:
Add constant.genie for zh-tw.

### DIFF
--- a/languages/zh-tw/constants.genie
+++ b/languages/zh-tw/constants.genie
@@ -1,0 +1,134 @@
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of ThingTalk
+//
+// Copyright 2017-2018 The Board of Trustees of the Leland Stanford Junior University
+//
+// Author: Giovanni Campagna <gcampagn@cs.stanford.edu>
+//
+// See COPYING for details
+
+{
+const assert = require('assert');
+
+const ThingTalk = require('thingtalk');
+const Ast = ThingTalk.Ast;
+const Type = ThingTalk.Type;
+const { clean } = require('../../lib/utils');
+
+// import the combinator library
+const C = require('../../lib/sentence-generator/ast_manip');
+}
+
+constant_Number = {
+    // `turking` is a flags that controls a group of grammar rules
+    // syntax is ?flag to enable the rule (or group of rules) only if the flag is set, !flag to disable the
+    // rule if the flag is set
+
+    !turking {
+    '0'    => new Ast.Value.Number(0);
+    '零' => new Ast.Value.Number(0);
+    '1'    => new Ast.Value.Number(1);
+    '一'  => new Ast.Value.Number(1);
+    }
+}
+
+constant_date_point = {
+    '現在'             => new Ast.Value.Date(null, '+', null);
+    '今天'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'day'), '+', null);
+    '昨天'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'day'), '-', Ast.Value.Measure(1, 'day'));
+    '這星期'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '+', null);
+    '這週'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '+', null);
+    '這禮拜'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '+', null);
+
+    !turking {
+    '明天'            => new Ast.Value.Date(Ast.DateEdge('end_of', 'day'), '+', null);
+    '今天以前'         => new Ast.Value.Date(Ast.DateEdge('end_of', 'day'), '+', null);
+    '這星期以前'       => new Ast.Value.Date(Ast.DateEdge('end_of', 'week'), '+', null);
+    '這週以前'         => new Ast.Value.Date(Ast.DateEdge('end_of', 'week'), '+', null);
+    '這禮拜以前' => new Ast.Value.Date(Ast.DateEdge('end_of', 'week'), '+', null);
+    '上星期'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '-', Ast.Value.Measure(1, 'week'));
+    '上週'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '-', Ast.Value.Measure(1, 'week'));
+    '上禮拜'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '-', Ast.Value.Measure(1, 'week'));
+    '這個月'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'mon'), '+', null);
+    '今年'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'year'), '+', null);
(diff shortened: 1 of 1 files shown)


languages/zh-tw/constants.genie
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+constant_date_point = {
+    '現在'             => new Ast.Value.Date(null, '+', null);
+    '今天'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'day'), '+', null);
+    '昨天'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'day'), '-', Ast.Value.Measure(1, 'day'));
+    '這星期'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '+', null);
+    '這週'             => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '+', null);
+    '這禮拜'           => new Ast.Value.Date(Ast.DateEdge('start_of', 'week'), '+', null);
```
gcampax: According to <https://corenlp.run>, these should be two words.
For speed, all NL fragments in the templates are tokenized/segmented ahead of time, hence you'll need to split them with a space.

johnnychhsu: I see. 
I was thinking about how it does Chinese segmentation lol
So the template we are going to do for device should be segment also right?

gcampax: No, primitive templates undergo segmentation by calling CoreNLP and should be written naturally.
Indeed, if you look at the `dataset.tt` that comes out of calling `genie download-dataset`, you'll see each example has `#_[utterances]` and `#_[preprocessed]` - developers provide the former in Thingpedia, and the latter is generated automatically by tokenizing `utterances`.

languages/zh-tw/constants.genie
```diff
@@ -0,0 +1,134 @@
...
+}
+
+constant_Measure_C = {
+    !turking {
+        num:constant_Number 'c' => C.addUnit('C', num);
+        num:constant_Number '攝氏' => C.addUnit('C', num);
+    }
+    num:constant_Number 'f' => C.addUnit('F', num);
+    num:constant_Number '華氏' => C.addUnit('F', num);
+    num:constant_Number '度' => C.addUnit('TB', num);
```
gcampax: Wrong unit here (and ugh it's broken in English too...)

Also, the English is written from an American point of view (using imperial units) but make sure the Chinese uses metric where ambiguous.

languages/zh-tw/constants.genie
```diff
@@ -0,0 +1,134 @@
...
+    !turking num:constant_Number 'm/s' => C.addUnit('mps', num);
+    num:constant_Number 'kph' => C.addUnit('kmph', num);
+    num:constant_Number 'miles per hour' => C.addUnit('mph', num);
+    num:constant_Number 'kilometers per hour' => C.addUnit('kmph', num);
+    num:constant_Number 'km/h' => C.addUnit('kmph', num);
+}
+
+constant_Location = {
+    '這裡'                    => new Ast.Value.Location(Ast.Location.Relative('current_location'));
+    !turking '我 在 哪裡' => new Ast.Value.Location(Ast.Location.Relative('current_location'));
```
gcampax: This is a question form, correct? The English template `where I am now` was meant for a relative clause used a parameter, as in `call an uber from where I am now to ...` or `find me a restaurant around where I am now`

languages/zh-tw/constants.genie
```diff
@@ -0,0 +1,134 @@
...
+constant_Date = {
+    !turking {
+    duration:constant_Measure_ms '以內' => new Ast.Value.Date(null, '+', duration);
+    duration:constant_Measure_ms '之內' => new Ast.Value.Date(null, '+', duration);
+    duration:constant_Measure_ms '以前' => new Ast.Value.Date(null, '-', duration);
+    duration:constant_Measure_ms '之前' => new Ast.Value.Date(null, '-', duration);
+    duration:constant_Measure_ms pt:constant_date_point '以後' => new Ast.Value.Date(pt.value, '+', duration);
+    duration:constant_Measure_ms pt:constant_date_point '之後' => new Ast.Value.Date(pt.value, '+', duration);
+    duration:constant_Measure_ms pt:constant_date_point '以前' => new Ast.Value.Date(pt.value, '-', duration);
+    duration:constant_Measure_ms pt:constant_date_point '之前' => new Ast.Value.Date(pt.value, '-', duration);
```
gcampax: I noticed that you're relying on the default definition of `constant_Measure_ms`. In English, that maps to a single `DURATION` token, because all the heavy lifting of understanding durations and time spans is done by CoreNLP through almond-tokenizer.
According to <https://corenlp.run>, this does not seem to be happening for Chinese. 
It will be good to understand what's going on, and what named entities we should rely on almond-tokenizer to understand, and what we'll let the neural parser learn.

(For English, NER is applied to locations, durations, times, dates and numbers, as well as a couple of language-agnostic regular expression patterns such as URLs, emails and hashtags; there is also an Americentric phone number recognizer that might need adaptation)

johnnychhsu: Thanks for the explanation!
I knew this problem (kph, km/h ...) when translate constant, but have no good idea then though, planned to work on this today.

BTW I push to this branch every time I have some updates (not finished yet) in case there are too many stuff to review. Is this OK for you?

gcampax: Yeah that works, just tell me when you're ready to review. Although I went through it a couple times already so I don't expect there will be much left to comment on.

languages/zh-tw/constants.genie
johnnychhsu: I see. Thank you!

languages/zh-tw/constants.genie
johnnychhsu: Yeah that is question form. 
Misunderstanding this, fix.

languages/zh-tw/constants.genie
johnnychhsu: OK I am not quite sure what TB is here...

languages/zh-tw/constants.genie
johnnychhsu: Usually we say '時速 x 公里', meaning x km/h. This is what we use for measuring speed, most of the time we only use this (almost all the cases).

johnnychhsu: There is no duration for Chinese in corenlp. 
I think we can use constant_date_point for concept like last week, next month, two weeks ago, after one year ... etc. As for duration, corenlp does not support now.

languages/zh-tw/constants.genie
gcampax: if that just means "degrees" it should be the default measure unit for temperatures: C in Chinese and F in (American) English
The TB there it's a bug in the current code

languages/zh-tw/constants.genie
gcampax: Assuming durations in Chinese are simpler than English (where you have weird stuff like a "fortnight") you can also add duration support by defining
```
constant_Measure_ms = {
  num:constant_Number "unit 1" => addUnit(num, '...');
  num:constant_Number "unit 2" => addUnit(num, '...');
  ...
```
like we for other non-time units.

languages/zh-tw/constants.genie
johnnychhsu: Got it. I delete it here.

gcampax:
Lets get this in so you can move on...

johnnychhsu:
Sure!

PR #11, opened 2019-02-27
add filters.genie for Chinese

johnnychhsu:
translate filters.genie for Chinese

### DIFF
--- a/languages/zh-tw/filters.genie
+++ b/languages/zh-tw/filters.genie
@@ -24,6 +24,9 @@ atom_filter = {
     t1:constant_Time '之前' => C.timeGetPredicate($options, null, t1);
     t2:constant_Time '之後' => C.timeGetPredicate($options, t2, null);
     t1:constant_Time '和' t2:constant_Time '之間' => C.timeGetPredicate($options, t1, t2);
+    '從' t1:constant_Time '到' t2:constant_Time => C.timeGetPredicate($options, t1, t2);
+    '從' t1:constant_Time '到' t2:constant_Time '之間' => C.timeGetPredicate($options, t1, t2);
+    '從' t1:constant_Time '開始 到' t2:constant_Time '結束' => C.timeGetPredicate($options, t1, t2);
 
     '我 在' loc:constant_Location     => C.locationGetPredicate($options, loc);
     '我 不 在' loc:constant_Location   => C.locationGetPredicate($options, loc, true);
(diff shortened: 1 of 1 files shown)


johnnychhsu:
@gcampax 
I need something like timeGetPredicate for constant_Date, should I add in ast_manip?
For example, d1:constant_Date '和' d2:constant_Date '之間' => C.dateGetPredicate($options, d1, d2);

gcampax:
You can add the helpers and new code to ast_manip yeah.
The question is, why do you need it?
Get predicates are used to add extra conditions to programs (or, more commonly, access control policies). A condition on dates seems odd? What's your use case?

johnnychhsu:
for example, if I want my blood pressure records between 2019/01/01~2019/02/25, I will say 'Give me my BP between 2019/01/01~2019/02/25', then I need that.

gcampax:
A get predicate is a filter on the current date when the code is evaluated, not on the date of the record. What you want is a range filter of type Date.

johnnychhsu:
I see. Thank you!

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+        if (v1.equals(v2)) // can happen with constants (now, 0, 1, etc.)
+            return null;
+        if (v1.isVarRef && v1.constNumber !== undefined && v2.isVarRef && v2.constNumber !== undefined &&
+            v1.constNumber + 1 !== v2.constNumber) // optimization: avoid CONST_X CONST_Y with X + 1 != Y earlier (before the NN catches it)
+            return null;
+        return [v1, v2];
+    };
+    v1:constant_Date ('和' | '跟') v2:constant_Date '之間'=> {
+        if (!v1.getType().equals(v2.getType()))
+            return null;
```
gcampax: This will never trigger, because range_filter only pairs the range to a numeric (number, currency, measure) parameter, not a date parameter.

Also, the `v1.getType().equals( v2.getType())` check is redundant because you're asking for Date constants through the grammar non-terminal.

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+const { clean } = require('../../lib/utils');
+
+// import the combinator library
+const C = require('../../lib/sentence-generator/ast_manip');
+}
+
+atom_filter = {
+    t1:constant_Time '之前' => C.timeGetPredicate($options, null, t1);
+    t2:constant_Time '之後' => C.timeGetPredicate($options, t2, null);
+    t1:constant_Time '和' t2:constant_Time '之間' => C.timeGetPredicate($options, t1, t2);
```
rayslxu: maybe also add "从...到..." / "从...开始到...结束"

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+
+    '我 在' loc:constant_Location     => C.locationGetPredicate($options, loc);
+    '我 不 在' loc:constant_Location   => C.locationGetPredicate($options, loc, true);
+
+    // What is projection here?
+    // proj:projection_Any ('is' | 'is exactly' | 'is equal to') x:constant_Any        => C.makeGetPredicate(proj, '==', x);
+    // proj:projection_Any ('is not' | 'is n\'t' | 'is different than') x:constant_Any => C.makeGetPredicate(proj, '==', x, true);
+
+    p:out_param_Any ('等於' | '相等於') x:constant_Any                  => C.makeFilter($options, p, '==', x);
```
rayslxu: I think a more natural way will be "是" or "为" for `==`, e.g., "温度为30度" instead of "温度等于30度"

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+    p:out_param_Any ('和' | '跟') x:constant_Any ('一樣' | '相同')      => C.makeFilter($options, p, '==', x);
+    p:out_param_Any '不等於' x:constant_Any                            => C.makeFilter($options, p, '==', x, true);
```
rayslxu: same here

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+    p:out_param_Any ('和' | '跟') x:constant_Any ('不一樣' | '不 相同')   => C.makeFilter($options, p, '==', x, true);
+
+    p:the_out_param_Numeric '比' x:constant_Numeric ('大' | '多') => C.makeFilter($options, p, '>=', x);
+    p:the_out_param_Numeric '比' x:constant_Numeric ('小' | '少') => C.makeFilter($options, p, '<=', x);
+
```
rayslxu: maybe "大于" / "小于"?

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+    // I need something like timeGetPredicate for constant_Date, should I add in ast_manip?
+    // For example, d1:constant_Date '和' d2:constant_Date '之間' => C.dateGetPredicate($options, d1, d2);
+
+    !turking {
+        p:the_out_param_Date '在' x:constant_Date '之後' => C.makeFilter($options, p, '>=', x);
+        p:the_out_param_Date '在' x:constant_Date '之前' => C.makeFilter($options, p, '<=', x);
+    }
+
+    p:the_out_param_Array__Any ('有' | '包含' | '擁有' | '含有' | '存在') x:constant_Any          => C.makeFilter($options, p, 'contains', x);
+    p:the_out_param_Array__Any ('沒有' | '不 包含' | '沒 擁有' | '不 含有' | '不 存在') x:constant_Any  => C.makeFilter($options, p, 'contains', x, true);
```
rayslxu: for "存在", I feel like it should be reversed: `x:constant_Any ('不 存在 于 |  不 在') p:the_out_param_Array__Any '中'`

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+
+    !turking range_filter;
+    //!turking f:either_filter => f;
+}
+
+edge_filter = {
+    !turking '當' p:out_param_Any ('跟' | '和') x:constant_Any ('一樣' | '相同') ('時' | '的 時候') => C.makeFilter($options, p, '==', x);
```
rayslxu: In your templates, the only difference between edge and non-edge filters is the word '當'. I would suggest using more obvious terms like "变成"
e.g.: `!turking '當' p:out_param_Any ('变成') x:constant_Any ('時' | '的 時候')`?

languages/zh-tw/filters.genie
```diff
@@ -0,0 +1,147 @@
...
+    param:the_out_param_Numeric '在' range:range => {
+        const [v1, v2] = range;
+        return new Ast.BooleanExpression.And([
+            Ast.BooleanExpression.Atom(param.name, '>=', v1),
+            Ast.BooleanExpression.Atom(param.name, '<=', v2)
+        ]);
+    };
+}
+
+with_filter = {
```
rayslxu: Actually, what's the equivalent of 'with a equals to be' in Chinese? In English it looks like "get the email with subject equals bla", in Chinese, it seems we only say ”获取标题为bla的邮件“, we don't think we put the conditions at the end of the sentence.

languages/zh-tw/filters.genie
johnnychhsu: Sure!

languages/zh-tw/filters.genie
johnnychhsu: yeah I think so. Thank you!

languages/zh-tw/filters.genie
johnnychhsu: Sure.

languages/zh-tw/filters.genie
johnnychhsu: yeah i agree. Thank you!

languages/zh-tw/filters.genie
johnnychhsu: I add something like the structure. I think we can still keep the original one, how do you think?

languages/zh-tw/filters.genie
johnnychhsu: That's right. we don't.

johnnychhsu:
> I add some comments based on Chinese in Mainland China. (things might be slightly different in Taiwan, I don't know)

Thank you! I modify some according to your review.

PR #12, opened 2019-02-27
add timers.genie for Chinese

johnnychhsu:
translate and modify timers.genie for Chinese


johnnychhsu:
I am done with this now.

PR #14, opened 2019-03-05
Revise Readme

elvisyjlin:
Revise the generation instruction in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,6 @@ genie generate --locale en --template template.genie --thingpedia thingpedia.jso
 The `--template` flag can be used to point to a template file definining the construct templates,
 in Genie language. E.g. `languages/en/thingtalk.genie` is the one for English sentence synthesis.
 Multiple `--template` flags can be used to load multiple template files.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,13 +76,14 @@ Please refer to the almond-tokenizer documentation for details.
...
 ```
 
 The `--template` flag can be used to point to a template file definining the construct templates,
-in Genie language. Multiple `--template` flags
-can be used to load multiple template files. If Genie has write permissions on the directory
-containing the template files, the files will be automatically compiled to JS and cached.
+in Genie language. E.g. `languages/en/thingtalk.genie` is the one for English sentence synthesis.
+Multiple `--template` flags can be used to load multiple template files.
+If Genie has write permissions on the directory containing the template files,
+the files will be automatically compiled to JS and cached.
```
gcampax: We should probably say that we don't cache any more...

elvisyjlin: I'll remove the last line "If Genie ... and cached."

PR #17, opened 2019-03-05
add aggregation/genie for Chinese

johnnychhsu:
add aggregation/genie for Chinese


johnnychhsu:
For average case, why one is set turking but the other one is not?
https://github.com/Stanford-Mobisocial-IoT-Lab/genie-toolkit/blob/master/languages/en/aggregation.genie#L47

gcampax:
"mean" is somewhat of a technical word, so we tought it would be unsafe to show to turkers, but it's a fine word to include in the synthetic set.

johnnychhsu:
I see. I am done with this.

rayslxu:
oh, btw, probably change the license & author at the top? don't give all the credit of your work to Giovanni, right? :)

johnnychhsu:
yeah XD Thank you!!

PR #19, opened 2019-03-06
add thingtalk.genie for Chinese

johnnychhsu:
Start to modify for Chinese. Not done yet.

### DIFF
--- a/languages/zh-tw/thingtalk.genie
+++ b/languages/zh-tw/thingtalk.genie
@@ -1,0 +1,454 @@
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of ThingTalk
+//
+// Copyright 2017-2018 The Board of Trustees of the Leland Stanford Junior University
+//
+// Author: Giovanni Campagna <gcampagn@cs.stanford.edu>
+//
+// See COPYING for details
+
+{
+const assert = require('assert');
+
+const ThingTalk = require('thingtalk');
+const Ast = ThingTalk.Ast;
+const Type = ThingTalk.Type;
+
+// import the combinator library
+const C = require('../../lib/sentence-generator/ast_manip');
+
+}
+
+import './constants';
+import './timers';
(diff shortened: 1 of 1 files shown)


johnnychhsu:
Currently done.

languages/zh-tw/thingtalk.genie
```diff
@@ -0,0 +1,390 @@
...
+    table:complete_get_command ('然後' | '接著') action:thingpedia_action if !complete => new Ast.Statement.Command(table, [action]);
+    table:complete_get_command '，' action:thingpedia_action if !complete => new Ast.Statement.Command(table, [action]);
+}
+
+complete_forward_get_do_command = {
+    forward_get_do_command if complete;
+}
+
+backward_get_do_command = {
+    action:thingpedia_action '當' ('拿到' | '取得' | '得到') table:complete_get_command ('時' | '的 時候') if !complete => new Ast.Statement.Command(table, [action]);
```
rayslxu: so far, I think you have been following what we have in English line-by-line. 
But Chinese is such a different language, we should have different constructs. 

for example, this `backward_get_do_command` does not make sense as far as I know. we don't say "发消息当下雨的时候", we only go forward, never backward. 
Of course, we will learn more by eyeballing through paraphrase after the synthetic sentences are generated. But I just want to make sure you guys keep this in mind, instead of copying whatever we have done in English.

johnnychhsu: Thank you!
I knew that Chinese would be quire different in some cases, so I remove all `with_....`. But the backward case, sometimes I do talk like that, although not very often, but it did happen. For example, '記得拿走喔！你要回去的時候'
But if that doesn't make sense to you, I think we can remove it.
What do you think?

rayslxu: Hmm, maybe. let's keep it then.

johnnychhsu: Got it!
Let's see what we will need after some synthetic experiments.

PR #23, opened 2019-03-17
Revision for zh-tw

elvisyjlin:
These files had some typos so that genie compiler reported Syntax Errors. Fixed them.


elvisyjlin:
In the second commit, a post processing function postprocessSynthetic() is added in `i18n/tranditional-chinese`. Genie generate will fail without it.

PR #35, opened 2019-04-15
Add enum value translation for Chinese

johnnychhsu:
Modify postprocessSynthetic() in tranditional-chinese.js.  Adding translation code


johnnychhsu:
add tests for traditional-chinese.js postprocessSynthetic().

PR #39, opened 2019-04-17
Fix the Parameter Template for Traditional Chinese

elvisyjlin:
This is the right revision. Please review it. Thank you!


gcampax:
Are all the placeholders single words according to corenlp?

elvisyjlin:
Oh, they should be segmented. I'll revise them.

PR #40, opened 2019-04-19
Add Document for i18n and ThingTalk Dataset Segmentator

elvisyjlin:


elvisyjlin:
Wait, I have to add test script for it.

gcampax:
This pull request **introduces 4 alerts** when merging 59de131162fb005a8a3252172f40435065b55984 into ff1513b10e6aa07813cc72d03aa4af72ad7b3f63 - [view on LGTM.com](https://lgtm.com/projects/g/stanford-oval/genie-toolkit/rev/pr-21f31d65f39dc7d16a3330d003b895327542e789)

**new alerts:**

* 3 for Unused variable, import, function or class
* 1 for Useless assignment to local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

elvisyjlin:
I realized that the testing procedure involves connecting to a almond tokenizer that supports Chinese. Is it feasible?

gcampax:
This pull request **introduces 1 alert** when merging 3ad2744aca60e03bbefbf9fc7355e3194fd5c736 into ff1513b10e6aa07813cc72d03aa4af72ad7b3f63 - [view on LGTM.com](https://lgtm.com/projects/g/stanford-oval/genie-toolkit/rev/pr-59ed7acdd7876fc8bf63edd652c5d83ea9c73614)

**new alerts:**

* 1 for Useless assignment to local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

gcampax:
> I realized that the testing procedure involves connecting to a almond tokenizer that supports Chinese. Is it feasible?

Yes. One needs to run to run almond-tokenizer locally, then set `GENIE_USE_TOKENIZER=local`.
The setup instructions include this.

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+        this._language = '';
+        this._tokenizer = TokenizerService.get(process.env.GENIE_USE_TOKENIZER, true);
+        this._num_examples = 0;
+        this._bar = null;
+        this._i = 0;
+        
+        if (this.locale.toLowerCase() == 'zh-tw') {
+            this._opencc_t2s = new OpenCC('t2s.json');
+            this._opencc_s2t = new OpenCC('s2t.json');
+        }
```
gcampax: This code is language specific, hence it should be in lib/i18n/traditional-chinese.js

Also, I'm assuming the same will be needed at inference time (i.e. when actually taking a sentence from the user)?
If so, it needs to be public API of Genie. Or ideally, it would be in almond-tokenizer.

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+//
+// Copyright 2019 National Taiwan University
+//
+// Author: Elvis Yu-Jing Lin <r06922068@ntu.edu.tw> <elvisyjlin@gmail.com>
+//
+// See COPYING for details
+"use strict";
+
+const stream = require('stream');
+const ProgressBar = require('progress');
```
gcampax: We already have a ProgressBar implementation in the library, do you need a new one?

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+        if (this.locale.toLowerCase() == 'zh-tw') {
+            this._opencc_t2s = new OpenCC('t2s.json');
+            this._opencc_s2t = new OpenCC('s2t.json');
+        }
+    }
+
+    async _transform(ex, encoding, callback) {
+        if (ex.type == 'meta') {
+            this._language = ex.data.language;
+            this._bar = new ProgressBar('[ :bar ] :current/:total', { total: ex.data.num_examples });
```
gcampax: Streams should not generate standard error output, that's the job of the wrapping tool code (see e.g. tool/train.js). The reason is that streams are composable and operate in interleaved fashion, so you could end up with two progress bars from two streams displaying at the same time.

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+            this._i += 1;
+            this._bar.tick({ current: this._i });
+        }
+        callback(null, ex);
+    }
+    
+    async preprocessAll(utterances, id) {
+        const promises = await utterances.map(async utterance => {
+            if (this.locale.toLowerCase() == 'zh-tw')
+                utterance = this._opencc_t2s.convertSync(utterance)
```
gcampax: Is there a non-sync version of this API?

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+                throw e;
+            }
+            this._i += 1;
+            this._bar.tick({ current: this._i });
+        }
+        callback(null, ex);
+    }
+    
+    async preprocessAll(utterances, id) {
+        const promises = await utterances.map(async utterance => {
```
gcampax: By convention, our style is to just pass the result of `.map` to `Promise.all()` unless there is a need to modify the array. Also, you should not `await` the result of `.map`: it's an Array of Promise, not a Promise.

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+            if (this.locale.toLowerCase() == 'zh-tw')
+                utterance = this._opencc_t2s.convertSync(utterance)
+            let preprocessed = await tokenizeExample(this._tokenizer, utterance, id, this._language);
+            if (this.locale.toLowerCase() == 'zh-tw')
+                preprocessed = this._opencc_s2t.convertSync(preprocessed)
+            return preprocessed;
+        });
+        const preprocessedAll = await Promise.all(promises);
```
gcampax: Don't await here, just return the result of `Promise.all` (see the [eslint rule](https://github.com/eslint/eslint/blob/master/docs/rules/no-return-await.md) for explanation).

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+                    language: this._dataset.language,
+                    annotations: this._dataset.annotations,
+                    num_examples: this._dataset.examples.length
+                }
+            });
+        } else {
+            this.push({
+                type: 'example',
+                data: this._dataset.examples[this._i]
+            });
```
gcampax: Pushing two different things in the same stream smells of bad design. Maybe streams are not the right abstraction here? After all, your streams don't stream at all, everything is cached in memory at once.

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+        
+        if (this._options.debug) {
+            console.log('Loaded ' + this._dataset.examples.length + ' templates');
+        }
+    }
+    
+    async _loadDataset() {
+        const code = await this._tpClient.getAllExamples();
+        const parsed = await Grammar.parse(code);
+        return parsed.datasets[0];
```
gcampax: No typecheck? (Here or later when you push the example)

tool/thingtalk-dataset.js
```diff
@@ -0,0 +1,312 @@
...
+    ThingTalkDatasetPreprocessor,
+    ThingTalkDatasetReader,
+    ThingTalkDatasetWriter
+};
+
+/* The following part is modified from Almond-Cloud
+ * https://github.com/stanford-oval/almond-cloud/blob/master/util/tokenize.js
+ * https://github.com/stanford-oval/almond-cloud/blob/master/util/validation.js
+ */
+const PARAM_REGEX = /\$(?:\$|([a-zA-Z0-9_]+(?![a-zA-Z0-9_]))|{([a-zA-Z0-9_]+)(?::([a-zA-Z0-9_-]+))?})/;
```
gcampax: It would be good to have all the copied over code into a library, so that almond-cloud can switch to the copy of the code in Genie.
Also, if you copy code you must either add a copyright notice, or just add shared copyright at the top.

tool/thingtalk-dataset.js
elvisyjlin: Yes, we also need it when a user type text in Traditional Chinese. However, the Almond tokenizer API need to be changed since it takes "zh" as Chinese, regardless of the Simplified one or the Traditional one.

tool/thingtalk-dataset.js
elvisyjlin: Where is our implementation?

elvisyjlin: Sorry for my stupid question, I found it in `lib/progress_bar`. Thanks!

tool/thingtalk-dataset.js
elvisyjlin: Yes, there is. I'll use `convert()` instead of `convertSync()`.

tool/thingtalk-dataset.js
elvisyjlin: 1. I put type and data here because I need to hold the meta data of the dataset as well to reconstruct dataset.tt. I search before but nodejs does not have streaming with meta data officially.
2. It's not streaming because dataset.prettyprint() is not streaming.

So...

Should I either 1) not to user streaming when processing tt dataset, or 2) write a streaming version of prettyprint()?

tool/thingtalk-dataset.js
elvisyjlin: Okay I'll put them in a library and write the correct copyright. Thanks.

tool/thingtalk-dataset.js
elvisyjlin: Got it. I noticed that I've awaited it here `ex.data.preprocessed = await this.preprocessAll(ex.data.utterances, ex.data.id);`.

tool/thingtalk-dataset.js
gcampax: Adding support for country codes in locale tags should be easy to do in almond-tokenizer, and you can then wrap the client-side code appropriately to do the right split, or just pass the unmodified locale tag to almond-tokenizer.

tool/thingtalk-dataset.js
gcampax: A streaming version of prettyprint() could be useful (for example in almond-cloud to implement APIs that return datasets more efficiently), but also not quite easy.
What I mean though, is that, if you need metadata, and metadata is not supported between streams, maybe nodejs streams/pipe are not the right abstraction and you should not use streams at all. Anyway this is more of style/readability comment than a hard requirement, maybe we'll clean it up in the future.

gcampax:
What's the status of this? Are you still working on it?

elvisyjlin:
I'm sick in bed this whole week... so I haven't done it. For the segmentation part, I'll remove OpenCC, use genie progress bar, get rid of stream/pipe. The OpenCC will be put in almond tokenizer.

elvisyjlin:
Updates:
* Removed OpenCC (the opencc part is moved to almond tokenizer)
* Removed streaming (the cleaner and preprocesser are replaced with a non-streaming version)

gcampax:
Let's get this over the finish line.

PR #41, opened 2019-04-25
FileThingpediaClient: remove legacy categories from thingpedia.json

gcampax:
Otherwise we can generate invalid examples


johnnychhsu:
test again and this solves the `online-account` error.

Issue #34, opened 2019-04-13
Add templates for the bookkeeping language


gcampax:
Bookkeeping commands (train, help, cancel, stop, go back, etc.) are necessary to any virtual assisstant, but so far training them has been annoying convoluted, requiring manual messing with the dataset from curl or mysql.
We should enable writing them quickly in template form, and then load on them.

All users of Genie will benefit, and it will be our first target language that is not ThingTalk.

PR #42, opened 2019-04-26
Add bookkeeping templates

gcampax:
This is an attempt to solve #34, using https://github.com/stanford-oval/thingtalk/pull/89

There is very little "templating" that goes into the bookkeeping commands, which are plain intents with no slots. Yet, I think writing the data for them in this fashion is easier than writing TSV files or using Train Almond.


gcampax:
@johnnychhsu @elvisyjlin this PR might of interest to you too; bookkeeping commands are special commands that the dialog agent needs, such as yes, no and cancel. You will need the equivalent in Chinese before you can use Almond for anything.

Issue #5, opened 2019-02-22
Datasets with context


gcampax:
Very soon, sentences will stop being interpreted in the vacuum, and will start being interpreted in context.
Hence, our dataset will grow from [id, sentence, program] to [id, context, sentence, program].

We need to:
- Define the language of contexts; probably a combination of ThingTalk, some stuff to talk about the current result, and the ability to concatenate multiple entries
- Extend file formats, parsers and stringifiers accordingly
- Add data augmentation passes to take non-contextual datasets and turn them into contextual datasets (naively, by sticking random contexts in front - maybe we can do better?)

Issue #6, opened 2019-02-22
Contextual templates


gcampax:
While templates probably won't be the end-all solution to data generation for contextual, at least we know some stuff can be represented in template form, for example filters after a policy request, follow up actions, or turning an immediate program into a monitor.

We should let developers express contextual templates, _with good sampling that does not bias the current context_.

This will be a follow up of #5

gcampax:
Fixed by #47

PR #43, opened 2019-05-02
First round of context templates

gcampax:
This is the initial implementation of context templates.

I'm looking mostly for feedback on the language, at this point.


gcampax:
This fixes #5, and is a first stab at #6

PR #65, opened 2019-07-23
Make use of the new canonical annotation for parameters

rayslxu:
This is backward compatible with string canonical, so it should be fine to merge


languages/en/filters.genie
```diff
@@ -180,3 +181,218 @@ range_with_filter = {
         ]);
     };
 }
+
+// npp filters
```
gcampax: I am ok with using abbreviations in the non-terminal name, but the comment is not very informative... Can you write down in the comment what each grammar category stands for, and maybe put the examples in?
I know everything is in the wiki, but people might not find the wiki at first...

languages/en/filters.genie
```diff
@@ -180,3 +181,218 @@ range_with_filter = {
...
+    p:out_param_apv x:constant_Any => {
+        return C.makeFilter($options, p, 'contains', x);
+    };
+}
+
+who_generic_filter = {
+    ('has' | 'got') p:out_param_npp x:constant_Any => {
+        const op = x.getType().isString ? '=~' : '==';
+        return C.makeFilter($options, p, op, x);
+    };
```
gcampax: has | got | has got ?

languages/en/filters.genie
```diff
@@ -180,3 +181,218 @@ range_with_filter = {
...
+
+    p:out_param_avp x:constant_Any => {
+        return C.makeFilter($options, p, 'contains', x);
+    };
+}
+
+
+// npi filters
+npi_filter = {
+    ('that' | '') 'is' 'the' p:out_param_npi x:constant_Any => {
```
gcampax: should we include a variant with "are" instead of "is" (here and elsewhere)?
maybe for robustness/completeness and not for turking? or as a postprocessing?

languages/en/who_questions.genie
```diff
@@ -0,0 +1,164 @@
...
+const ThingTalk = require('thingtalk');
+const Ast = ThingTalk.Ast;
+const Type = ThingTalk.Type;
+
+// import the combinator library
+const C = require('../../lib/sentence-generator/ast_manip');
+
+}
+
+
```
gcampax: We're going to need comments in this file...

languages/en/thingtalk.genie
```diff
@@ -321,13 +322,58 @@ $root = {
...
     ) => C.makeProgram(new Ast.Statement.Command(table, [C.notifyAction()]));
     ('hey almond' | 'please' | '') ('list' | 'enumerate') table:with_filtered_table => {
         if (!table.schema.is_list)
             return null;
         return C.makeProgram(new Ast.Statement.Command(table, [C.notifyAction()]));
     };
 
+    // now => get => notify (who questions)
+    !turking (
+          ('hey almond' | '') ('who' | 'which person' | 'which one' | 'which guy')  ('has' | 'got' ) table:who_npp_filtered_table ('' | '?')
```
gcampax: Should we go crazy with synonyms of "person" here? "individual", "human", ... ? "dude"?
Maybe we should only have "person", and use PPDB (or similar synonym database, eg WordNet) to replace during augmentation?

(interestingly, whether "guy" and "dude" are gendered are a matter of opinion and culture, and their use varies across English-speaking countries, which makes me think "guy" is not a great fit in the general templates and might be better as augmentation, where uncertainty can be accounted for)

Also, "which one" is kind of awkward? It seems to imply an existing set to choose from, and would likely be confusing when there is context.
Speaking of which... are you planning to incorporate context in any form? That is, if the previous statement produced a list of people, should the question become a refinement?

lib/sentence-generator/index.js
```diff
@@ -235,7 +261,7 @@ class SentenceGenerator extends events.EventEmitter {
                         {}
                     )]);
                 }
-                const pcanonical = ex.value.schema.getArgument(pname).canonical || clean(pname);
+                const pcanonical = ex.value.schema.getArgCanonical(pname);
```
gcampax: Is this new in ThingTalk? And do you need to bump the thingtalk dep to pass Travis?

lib/sentence-generator/index.js
```diff
@@ -249,10 +275,23 @@ class SentenceGenerator extends events.EventEmitter {
...
+            if (grammarCat === 'thingpedia_query') {
+                const human_entity_types = ['tt:contact', 'tt:contact_name', 'org.wikidata:human'];
+                if (ex.value.schema.hasArgument('id')) {
+                    let type = ex.value.schema.getArgument('id').type;
+                    if (type.isEntity && human_entity_types.includes(type.type)) {
+                        grammarCat = 'thingpedia_who_question';
+                        this._grammar.addRule(grammarCat, [''], $runtime.simpleCombine(() => ex.value));
+                        continue;
+                    }
+                }
```
gcampax: What's going on here? This code should be outside the `for (let preprocessed of ex.preprocessed)` loop right? Otherwise you add multiple rules? And you're adding rules with empty string for expansion?

Also, `tt:contact_name` is obsolete, we should use `tt:username` consistently to mean unresolved phone_number/email_address. And `tt:contact` is a type used by Comma to mean email/phone/matrix account, which doesn't seem relevant here?
Do you expect there will be a `@builtin.contacts()` that uses the OS address book (or some thingengine-core address book abstraction) and has an `id` parameter of type `tt:contact`?

languages/en/filters.genie
rayslxu: I thought about that. This will make the templates look way messier. I think postprocessing might be a better solution, which should include other grammar fixes, such as changing the pronoun and tense accordingly.

languages/en/thingtalk.genie
rayslxu: I removed "guy" and "which one", and added "individual". I agree that we should keep the more generic ones and use augmentation for others. 
No context. One problem at a time.

lib/sentence-generator/index.js
rayslxu: Fixed. `tt:contact` is still included in case we might want to use it in the future as in the example you provided.

PR #66, opened 2019-07-26
Wip/cardiology intervals

rickygv99:
I read through several instruction manuals / websites for at-home blood pressure measurements. All of them recommend taking your blood pressure twice a day so I targeted the multi-timer genie templates at 'twice a day' timers.

I then sent a survey to several groups of friends asking for natural language examples for taking a pill twice a day. I extended my templates to capture all the different variations they used in their phrasing.


rickygv99:
Thanks for the feedback! I changed the second-person references to first-person references. I'm working on refactoring the rules to use nonterminals so as to utilize recursive expansion rather than sampling. As per your recommendation, I also plan on introducing a couple more relative times for a couple of the rules (such as "night" / "afternoon" / "school [start/end]" / "work [start/end]").

Combined with these relative times, if we limit ourselves to capturing only time phrases (such as "before work" / "before school") which are relatively universal and frequently used in communication, rather than attempting to consider all time phrases personal to a user (such as "before soccer practice" / "before dance"), then I think the set of time phrases we need to capture will be bounded (and not that big) and doable to cover. In my survey I sent out, only a select few time phrases were used (such as "before meals" / "when you brush your teeth" / "before work"), making them good candidates to cover in our timer syntax.

rickygv99:
It would be interesting, as a future project, to also capture personal time phrases for the user. A virtual assistant should ideally include at least the core functions of a human assistant, such as keeping a calendar. We could allow the user to record calendar events on their virtual assistant and support commands referencing these events (e.g. If the user records an event of "Soccer practice on Mondays at 3 PM" then the virtual assistant will understand timer phrases such as "Before soccer practice"). After more thought, would be further interesting to integrate this, location data, and traffic data into a reminder system on Almond that sends event reminders to users at an early enough time that they arrive at their event on time, accounting for traffic between that event and their current location. Lots of cool things we could do with an integrated calendar system.

gcampax:
I merged this because today's [feature freeze release day](https://wiki.almond.stanford.edu/ReleasePlanning/1.6) but more work will be needed before we'll enable this feature in the default model. Luckily, soon you will be able to experiment with this in production by uploading custom Genie template packs.

PR #72, opened 2019-08-07
Add a topdown sentence generator

gcampax:
The topdown generator works by recursively sampling a new entry for each non-terminal in the rule.
This makes it lazy in its sampling, so it should be able to explore more of the space, without carrying over previous choices from one depth to the next.


gcampax:
Actually the new bottom-up generation with modified sampling is pretty decent, and it's way faster than top-down still, so no.

PR #88, opened 2019-09-06
lib/parallelize: implement back pressure

gcampax:
If all workers are busy (not reading from the stream), stop accepting
data from the input stream. This ensures that the input stream is advanced
at a controlled rate compared to the output stream, and allows measuring the
progress of the overall operation by measuring the progress of the
input stream.

Marking as draft because I haven't tested this, and also I'm not sure this approach makes sense.


gcampax:
This doesn't work well, it doesn't really do what's it's supposed to, and it makes the datasets non deterministic, which is a big no-no.

PR #98, opened 2019-10-20
Wip/cardiology intervals

greentfrapp:
Updated extended_timers in timers.genie and added them to remote_commands in thingtalk.genie. But the generated code wouldn't work yet because we still need to fix Timer in [thingengine-core/lib/apps/timers.js](https://github.com/stanford-oval/thingengine-core/blob/master/lib/apps/timers.js) to take frequency (see [relevant PR by Ricky](https://github.com/stanford-oval/thingengine-core/pull/86)).

### DIFF
--- a/languages/en/thingtalk.genie
+++ b/languages/en/thingtalk.genie
@@ -305,7 +305,7 @@ $root = {
         let stream = C.tableToStream(table, null);
         if (!stream)
             return null;
-        return C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction()]));
+        return C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction()]), principal);
     };
     !nofilter ('if' | 'when') filter:atom_filter 'in' table:complete_table ('for' | 'of') principal:constant_Entity__tt__username 'then' ('alert me' | 'tell me' | 'notify me' | 'let me know')  => {
         if (C.hasGetPredicate(filter))
@@ -320,7 +320,7 @@ $root = {
         let stream = C.tableToStream(table, null);
         if (!stream)
             return null;
-        return C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction()]));
+        return C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction()]), principal);
     };
     !nofilter ('alert me' | 'tell me' | 'notify me' | 'let me know') ('if' | 'when') filter:atom_filter 'in' table:complete_table => {
         if (C.hasGetPredicate(filter))
--- a/languages/en/thingtalk.genie
+++ b/languages/en/thingtalk.genie
@@ -292,6 +292,36 @@ $root = {
         // note that we intentionally don't use tableToStream here, because we don't want to introduce edge filters
         return C.makeProgram(new Ast.Statement.Rule(new Ast.Stream.Monitor(proj.table, null, proj.table.schema), [C.notifyAction()]));
     };
+    !nofilter ('alert me' | 'tell me' | 'notify me' | 'let me know') ('if' | 'when') filter:atom_filter 'in' table:complete_table ('for' | 'of') principal:constant_Entity__tt__username => {
+        if (C.hasGetPredicate(filter))
+            return null;
+        if (!table.schema.is_monitorable || !C.checkFilter(table, filter))
+            return null;
+        if ($options.flags.turking && table.schema.is_list)
+            return null;
+        table = C.addFilter(table, filter, $options);
+        if (!table)
+            return null;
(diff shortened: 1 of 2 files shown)


languages/en/thingtalk.genie
```diff
@@ -441,6 +441,14 @@ $root = {
...
         ) => C.makeProgram(new Ast.Statement.Command(table, [C.notifyAction('return')]), principal);
         ('tell' | 'command' | 'order' | 'request' | 'ask') principal:constant_Entity__tt__username 'to' ('let me know' | 'inform me' | 'notify me' | 'alert me') stream:stream => C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction('return')]), principal);
+
+        ?extended_timers {
+            (
+                ('tell' | 'command' | 'order' | 'request' | 'ask') principal:constant_Entity__tt__username 'to' action:thingpedia_action stream:timer
+                | ('tell' | 'command' | 'order' | 'request' | 'inform') principal:constant_Entity__tt__username 'that' ('he needs' | 'she needs' | 'i need him' | 'i need her') 'to' action:thingpedia_action stream:timer
+            ) if complete => C.makeProgram(new Ast.Statement.Rule(stream, [action]), principal);
+            ('tell' | 'command' | 'order' | 'request' | 'ask') principal:constant_Entity__tt__username 'to' ('let me know' | 'inform me' | 'notify me' | 'alert me') stream:timer => C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction('return')]), principal);
```
gcampax: the last one doesn't quite make sense: why would you go through a remote command to execute just the timer?

languages/en/timers.genie
```diff
@@ -347,6 +347,12 @@ when_brushing_teeth = {
   'after i brush my teeth' => 'after i brush my teeth';
 }
 
+on = {
+  'on' => 'on';
+  'when it\'s' => 'when it\'s';
```
gcampax: space between `it` and `'s`
(also, the difference between "it is" and "it's" should probably be handled in postprocessing)

languages/en/timers.genie
```diff
@@ -372,19 +378,38 @@ timer = {
 
         !turking {
             optional_once each interval:constant_Measure_ms => new Ast.Stream.Timer(Ast.Value.Date.now(), interval, null, TIMER_SCHEMA);
-            once_per_day => new Ast.Stream.AtTimer([TIME_MORNING], null, AT_TIMER_SCHEMA);
+            once_per_day => new Ast.Stream.Timer(Ast.Value.Date.now(), new Ast.Value.Measure(1, 'day'), null, TIMER_SCHEMA);
+            // twice_per_day and thrice_per_day still need to be AtTimer until we can explicitly set the frequency param in Timer
             twice_per_day => new Ast.Stream.AtTimer([TIME_MORNING, TIME_EVENING], null, AT_TIMER_SCHEMA);
             thrice_per_day => new Ast.Stream.AtTimer([TIME_MORNING, TIME_NOON, TIME_EVENING], null, AT_TIMER_SCHEMA);
```
gcampax: These two (twice_per_day / thrice_per_day) need to go

languages/en/thingtalk.genie
greentfrapp: You mean the one at line 450 right? Hmm we were wondering if this will allow both the sender and recipient to receive regular alerts? But I guess that's stretching it. Okay we'll drop that.

languages/en/timers.genie
greentfrapp: Okay will add the space.
Do you mean contractions are already handled or should we update the code for that somewhere? 
So I should keep 'when it \'s' and remove 'when it is', or the other way around?

languages/en/timers.genie
greentfrapp: We wanted to change them to Timer but couldn't set the freq parameter explicitly. The main problem is that the words 'twice' and 'thrice' wouldn't get mapped if we remove these.

languages/en/thingtalk.genie
```diff
@@ -292,6 +292,36 @@ $root = {
...
+            return null;
+        if ($options.flags.turking && table.schema.is_list)
+            return null;
+        table = C.addFilter(table, filter, $options);
+        if (!table)
+            return null;
+        let stream = C.tableToStream(table, null);
+        if (!stream)
+            return null;
+        return C.makeProgram(new Ast.Statement.Rule(stream, [C.notifyAction()]));
```
gcampax: You're ignoring principal here and in the next template (unless I'm mistaken)

gcampax:
Ok, let's merge this and do a round of training, we can play with templates again if needed.

PR #102, opened 2019-10-22
sentence-generator: fix getExamplesByKinds call

gcampax:
It should take an array of kinds, not a single one


gcampax:
Depends on https://github.com/stanford-oval/thingpedia-api/pull/43

PR #116, opened 2019-11-11
Simplified Chinese version of Genie

jack57lee:
A Simplified Chinese version of Genie, based on the Traditional Chinese one.

Since ThingTalk has changed a lot, the Traditional Chinese one has many bugs thus no longer works. I've resolved those bugs. I also fixed some expressions according to simplified Chinese grammar.

Besides, not all Genie features are included in this version, for example, who questions and argmin/argmax.


languages/zh-cn/aggregation.genie
```diff
@@ -0,0 +1,121 @@
...
+            return null;
+        if (!t.schema.is_list)
+            return null;
+
+        const newSchema = t.schema.filterArguments((arg, i) => arg.direction !== Ast.ArgDirection.OUT || arg.name === p.name);
+        const agg = new Ast.Table.Aggregation(t, p.name, 'avg', null, newSchema);
+        return new Ast.Table.Projection(agg, [p.name], newSchema);
+    };
+
+    !turking t:complete_table '的' p:out_param_Numeric '的 平均'=> {
```
rayslxu: Should be either "p 的平均值" or "平均 p" right?

jack57lee: Yes, I'll fix it.

PR #123, opened 2020-01-11
Abstractions over multiple target languages, try 2

gcampax:
No multiwoz related code in this branch, just code cleanups and the script to generate dialogues (that can run for ThingTalk too)


languages/thingtalk/load-thingpedia.js
```diff
@@ -31,7 +31,9 @@ class ThingpediaLoader {
         this._grammar = grammar;
 
         this._tpClient = options.thingpediaClient;
-        this._schemas = options.schemaRetriever || new SchemaRetriever(this._tpClient, null, !options.debug);
+        if (!options.schemaRetriever)
+            options.schemaRetriever = new SchemaRetriever(this._tpClient, null, !options.debug);
+        this._schemas = options.schemaRetriever;
```
rayslxu: why this change?

gcampax:
Commit 98e7e8 modifies the data structures used by the sentence generator. Before, everything was a in JS objects used as hash tables, keyed by non-terminal name. After, it's all arrays, which are accessed by precomputed index. This avoids hashing strings and replaces hashtable lookups with a faster array lookups.
The biggest win should be the `charts` data structure, which holds the intermediate derivations at the previous depth, and is accessed in the hot inner loop where we expand each rule.
I don't have performance numbers before and after, but unscientifically I was seeing about a 20-30% improvement. Not much, but not small either.

languages/thingtalk/load-thingpedia.js
gcampax: Before this PR, schemaRetriever was usually provided by the API caller (ie tool/generate.js, or almond-cloud). Now it is not, because we want to cut the number of files that `require('thingtalk')`. But we still don't want to create too many schema retrievers, or we lose the caching they provide and hit the Thingpedia API too often.
So with this change, if the schema retriever is created lazily, it is also stored in `options` for other code to use (specifically, for languages/thingtalk.js), and we only create it once.

PR #127, opened 2020-02-22
Map 'degrees' to defaultTemperature unit

greentfrapp:
Update from previous version that mapped 'degrees' to 'Farenheit


gcampax:
I created the `stable-0-5` branch, can you cherry-pick/rebase your patch on top of that branch and update this PR? You should also do a `yarn upgrade` to pick up the latest version of ThingTalk.

PR #132, opened 2020-03-09
Added option to use fine-tuned bert

s-jse:


tool/lib/bert-canonical-generator.py
```diff
@@ -7,22 +7,25 @@
...
 
 class BertLM:
-    def __init__(self, domain, examples, mask, k):
+    def __init__(self, domain, examples, mask, k, model_name_or_path='bert-large-uncased', is_paraphraser=False):
         """
         :param domain: an object contains the canonical form and paths to parameters for each table in the domain
         :param examples: an object of examples for each grammar category of each property of each table
         :param mask: a boolean indicates if we do masking before prediction
         :param k: number of top candidates to return per example
         """
```
rayslxu: could you add the comment as I did?

tool/lib/bert-canonical-generator.py
```diff
@@ -214,16 +234,23 @@ def construct_canonical(query, masks, current_index, replacement):
                         type=int,
                         default=500,
                         help='top-k candidates to return when generating adjectives')
+    parser.add_argument('--model-name-or-path',
+                        type=str,
+                        default='bert-large-uncased',
+                        help='The name of the model (e.g. bert-large-uncased) or the path to the directory where the model is saved.')
+    parser.add_argument('--is-paraphraser',
+                        action='store_true',
+                        help='If the model has been trained on a paraphrasing corpus')
```
rayslxu: set the default from here, instead of on the function, so we don't need to pass additional options most of the time.

s-jse: The defaults are set in both places (`--is-paraphraser` is False by default when not specified), but I can remove them from the function.

rayslxu: oh, i see. that's fine. just add the docstring, then let's merge this, so i can run the new experiments.

s-jse:
Please do a "squash and merge" instead of merge.

PR #144, opened 2020-03-28
More changes from schema2qa

rayslxu:
- Fix removing name from schemas
- Add the option to not always generate base form canonical
- Use type info for base form canonical if absent
- Migrate to new annotations, no more npps


tool/lib/webqa-canonical-generator.js
```diff
@@ -97,9 +105,23 @@ class CanonicalGenerator {
...
+                if (arg.metadata.canonical.base && !arg.metadata.canonical.property)
+                    arg.metadata.canonical.property = [...arg.metadata.canonical.base];
+
+                // if property is missing, try to use entity type info
+                if (!('property' in arg.metadata.canonical)) {
+                    let typestr = entityTypeToString(query.getArgType(arg.name));
+                    // only apply this if the type is unique
+                    if (typestr && typeCounts[typestr] === 1) {
+                        let base = typestr.substring(typestr.indexOf(':') + 1)
+                            .replace(/_/, ' ')
```
gcampax: Should probably use "clean()" from Utils instead of this replace, as that handles different casing conventions and does a few more fixes.

tool/lib/webqa-canonical-generator.js
```diff
@@ -20,6 +20,17 @@ const ANNOTATED_PROPERTIES = [
     'geo', 'address.streetAddress', 'address.addressCountry', 'address.addressRegion', 'address.addressLocality'
 ];
 
+
+// turn entity type into string
+function entityTypeToString(type) {
+    if (type.isArray)
+        return entityTypeToString(type.elem);
```
gcampax: It took me an extra second to understand what this function does. The name is not great, because it doesn't turn the type into a string, it find the entity type name out of a type. So maybe `typeToEntityType` or something like that? Or a better comment that doesn't just repeat the function name...

tool/lib/webqa-canonical-generator.js
```diff
@@ -97,9 +105,23 @@ class CanonicalGenerator {
...
-                if (arg.metadata.canonical.base && !arg.metadata.canonical.npp)
-                    arg.metadata.canonical.npp = [...arg.metadata.canonical.base];
+                // copy base canonical if property canonical is missing
+                if (arg.metadata.canonical.base && !arg.metadata.canonical.property)
+                    arg.metadata.canonical.property = [...arg.metadata.canonical.base];
+
+                // if property is missing, try to use entity type info
+                if (!('property' in arg.metadata.canonical)) {
+                    let typestr = entityTypeToString(query.getArgType(arg.name));
+                    // only apply this if the type is unique
```
gcampax: "is unique" here means only one property has this type right? Maybe spell it out...

tool/lib/webqa-canonical-generator.js
```diff
@@ -144,6 +166,21 @@ class CanonicalGenerator {
...
+        const schema = this.class.queries[qname];
+        const count = {};
+        for (let arg of schema.iterateArguments()) {
+            let typestr = entityTypeToString(schema.getArgType(arg.name));
+            if (!typestr)
+                continue;
+            if (typestr in count)
+                count[typestr] += 1;
+            else
+                count[typestr] = 1;
```
gcampax: You can write it more concisely as `count[typestr] = (count[typestr] || 0) + 1`
(not sure it's worth it)

tool/webqa-process-schemaorg.js
```diff
@@ -947,12 +883,21 @@ class SchemaProcessor {
...
+                // override name so we can apply a custom string_values annotation
+                const arg = new Ast.ArgumentDef(null, Ast.ArgDirection.OUT, 'name', Type.String, {
+                    nl: {},
+                    impl: {
+                        'org_schema_type': new Ast.Value.String('Text'),
+                        'filterable': new Ast.Value.Boolean(false) // no filter on name, if it has ner support, we'll generate prim for it
+                    }
+                });
+                recursiveAddStringValues(arg, this._prefix + typename + '_name');
+                args.push(arg);
```
gcampax: This is wrong and just sneaked in right? We don't want a `name` property.

tool/webqa-process-schemaorg.js
```diff
@@ -1053,6 +998,18 @@ module.exports = {
...
+            action: 'storeTrue',
+            help: `Always generate base canonical`,
+            defaultValue: true
+        });
+        parser.addArgument('--no-bad-base-canonical', {
+            nargs: 0,
+            action: 'storeFalse',
+            help: `Do not always generate base canonical`,
+            dest: `always_base_canonical`,
+        });
```
gcampax: By convention, if you have a pair of boolean flags, the second one is `--no-` the first one, so it's obvious they are paired.

tool/webqa-process-schemaorg.js
rayslxu: We still have the name in process step to allow us understand which table has name, and which id has ner support, we then remove name at the end of trimming

rayslxu: this is actually what the bug was

rayslxu:
@gcampax thanks for the detailed review, i just addressed all of them

tool/webqa-process-schemaorg.js
gcampax: Oh nice catch, but ugh this needs to be written down

PR #145, opened 2020-04-01
Inference mode dialogue templates

gcampax:
_This is not 1.8 material_

This PR introduces a new way to run the dialogue templates, in "inference mode". In this mode, we take exactly one context, and generate exactly one possible continuation. If this continuation contains only the agent sentence and agent dialogue state, it is suitable to choose as the policy and utterance in a real interactive agent.


gcampax:
@rayslxu I think this PR might be good for review now. The dialogue stuff is not complete yet, but I would rather avoid it grow indefinitely and only merge at the end. This is the branch that was used for the homework, with some of the fixes that came out of it, so it should be good.

gcampax:
(I thought I replied to this before)

Yeah you will need to update paraphrase and evaluation data to change the compute filter syntax. We should probably make that a normalization in ThingTalk...

PR #149, opened 2020-04-05
Various fixes discovered when preparing cs294s hw1

rayslxu:
Fix normalization
- do not remove name property even if it's not in the schema
- output both the data json and the meta json since they are both needed for the device package

Add whitelist annotation to device class
- genie will only load whitelisted functions if it's specified.

### DIFF
--- a/languages/thingtalk/load-thingpedia.js
+++ b/languages/thingtalk/load-thingpedia.js
@@ -64,7 +64,7 @@ class ThingpediaLoader {
         this.idQueries = new Map;
         this.compoundArrays = new Map;
         if (this._options.whiteList)
-            this.whiteList = this._options.whiteList.toLowerCase().split(',');
+            this.whiteList = this._options.whiteList.split(',');
         else
             this.whiteList = null;
 
@@ -595,7 +595,7 @@ class ThingpediaLoader {
     }
 
     async _loadFunction(functionDef) {
-        if (this.whiteList && !this.whiteList.includes(functionDef.name.toLowerCase()))
+        if (this.whiteList && !this.whiteList.includes(functionDef.name))
             return;
 
         let functionName = functionDef.class.kind + ':' + functionDef.name;
@@ -620,8 +620,8 @@ class ThingpediaLoader {
         let queries = Object.keys(classDef.queries);
         let actions = Object.keys(classDef.actions);
         if (whitelist && whitelist.length > 0) {
-            queries = queries.filter((name) => whitelist.includes(name.toLowerCase()));
-            actions = actions.filter((name) => whitelist.includes(name.toLowerCase()));
+            queries = queries.filter((name) => whitelist.includes(name));
+            actions = actions.filter((name) => whitelist.includes(name));
         }
 
         await Promise.all(queries.map((name) => classDef.queries[name]).map(this._loadFunction.bind(this)));
@@ -639,7 +639,7 @@ class ThingpediaLoader {
         if (this.idQueries.has(idEntity))
             return true;
 
-        if (this.whiteList && !this.whiteList.includes(suffix.toLowerCase()))
+        if (this.whiteList && !this.whiteList.includes(suffix))
             return false;
 
         let classDef;
--- a/languages/thingtalk/load-thingpedia.js
+++ b/languages/thingtalk/load-thingpedia.js
@@ -649,7 +649,10 @@ class ThingpediaLoader {
             // ignore if the class does not exist
             return false;
         }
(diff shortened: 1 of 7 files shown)


languages/thingtalk/load-thingpedia.js
```diff
@@ -615,8 +615,17 @@ class ThingpediaLoader {
...
+        const whitelist = classDef.getImplementationAnnotation('whitelist');
+        let queries = Object.keys(classDef.queries);
+        let actions = Object.keys(classDef.actions);
+        if (whitelist && whitelist.length > 0) {
+            queries = queries.filter((name) => whitelist.includes(name.toLowerCase()));
+            actions = actions.filter((name) => whitelist.includes(name.toLowerCase()));
+        }
+
+        await Promise.all(queries.map((name) => classDef.queries[name]).map(this._loadFunction.bind(this)));
+        await Promise.all(actions.map((name) => classDef.actions[name]).map(this._loadFunction.bind(this)));
```
gcampax, line 628: There are other parts that handle the whitelist right? Like the idType loading?

tool/webqa-process-schemaorg.js
```diff
@@ -962,7 +963,11 @@ class SchemaProcessor {
                 name: `${this._className ? this._className + ' in ' : ''}Schema.org`,
                 description: 'Scraped data from websites that support schema.org'
             },
-            impl: {}
+            impl: {
+                whitelist: new Ast.Value.Array(
+                    this._white_list.map((q) => new Ast.Value.String(q.trim().toLowerCase()))
```
gcampax, line 968: Why is it lowercased, and not case-sensitive? If it's developer-specified, you can match the case of the query.

tool/webqa-normalize-data.js
```diff
@@ -498,8 +499,10 @@ module.exports = {
             addHelp: true,
             description: "Normalize schema.org JSON+LD files to match their ThingTalk representation."
         });
-        parser.addArgument(['-o', '--output'], {
-            required: true,
+        parser.addArgument('--data-output', {
+            type: fs.createWriteStream
+        });
+        parser.addArgument('--meta-output', {
```
gcampax, line 505: It's somewhat unfortunate to change the command-line in a stable release.
It is also annoying to have a command line tool with two outputs, mostly because make really doesn't like that.

Also, is there a reason why these are not required outputs?

languages/thingtalk/load-thingpedia.js
rayslxu: yes, we have a global white list, if that is specified, all the other functions won't be loaded, including functions from other devices. we should probably remove that. but that also means changing Makefile....

tool/webqa-process-schemaorg.js
rayslxu: this is only because we have been using lowercase previously... i can remove that from all other scripts too.

tool/webqa-normalize-data.js
rayslxu: I don't want to create a separate script to do the job that has been done... any suggestions? 

they are not required, so i can choose to output only one in the Makefile :)

tool/webqa-normalize-data.js
```diff
@@ -513,10 +516,15 @@ module.exports = {
             nargs: '+',
             help: 'Input JSON+LD files to normalize. Multiple input files will be merged in one.'
         });
+        parser.addArgument('--class-name-prefix', {
+            required: false,
+            help: 'The prefix of the class name, defaults to org.schema',
+            defaultValue: 'org.schema'
+        });
```
gcampax, line 523: Why is it the class name prefix and not just the class name?

languages/thingtalk/load-thingpedia.js
gcampax: What I mean is, the local whitelist should also affect idType loading like the global whitelist does.

tool/webqa-normalize-data.js
rayslxu: hmm... true

rayslxu: addressed.

tool/webqa-process-schemaorg.js
rayslxu: addressed

languages/thingtalk/load-thingpedia.js
rayslxu: addressed

gcampax:
OK I think this is good now.

PR #152, opened 2020-04-07
Wip/multilanguage

Mehrad0711:
[ready for final review]

Hey, just trying to get your feedback on this before I commit more!
Let me know if it needs fixes/ changes so we can discuss it in the meantime...

Thanks ahead for the review :)


lib/i18n/default.js
```diff
@@ -110,6 +110,22 @@ function pluralize(noun) {
     return undefined;
 }
 
+function isGoodWord(word) {
+    // filter out words that cannot be in the dataset,
+    // because they would be either tokenized/preprocessed out or
+    // they are unlikely to be used with voice
+    // [Arabic_chars Arabic_Digits Farsi_supplements ...]
+    return undefined;
```
gcampax, line 118: Should always return `true` right? Absent any language-specific info, every word is a good word.

lib/dataset_augmenter.js
```diff
@@ -34,6 +34,7 @@ module.exports = class DatasetAugmenter extends Stream.Transform {
         this._schemas = schemaRetriever;
         this._paramReplacer = new ParameterReplacer(this._schemas, this._constProvider, {
             locale: options.locale,
+            paramLocale: options.paramLocale,
```
gcampax, line 37: What's this for?

lib/dataset_augmenter.js
```diff
@@ -44,6 +45,7 @@ module.exports = class DatasetAugmenter extends Stream.Transform {
             noQuoteExpandFactor: this._options.noQuoteExpandFactor,
             paraphrasingExpandFactor: this._options.paraphrasingExpandFactor,
             replaceLocations: this._options.replaceLocations,
+            replaceNumbers: this._options.replaceNumbers,
```
gcampax, line 48: This is going to be tricky to implement by the way, because NUMBER_ cannot always be replaced with SLOT_ tokens (which happens as a prereq to loading the program so you can typecheck & iterate the slots). It's mostly a problem around measurements, IIRC.

lib/i18n/persian.js
```diff
@@ -0,0 +1,115 @@
...
+    return undefined;
+}
+
+
+function isGoodWord(word) {
+    // filter out words that cannot be in the dataset,
+    // because they would be either tokenized/preprocessed out or
+    // they are unlikely to be used with voice
+    // [Arabic_chars Arabic_Digits Farsi_supplements ...]
+    return /^([\u0600-\u06ff\u0660-\u0669\uFB50–\uFDFF0-9-][\u0600-\u06ff\u0660-\u0669\uFB50–\uFDFF0-9.-]*|\u060C|\u061F)$/.test(word);
```
gcampax, line 78: Some punctuation is good, right? at least in English, we leave `?` and `!`

lib/i18n/persian.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function isGoodSentence(sentence) {
+    if (sentence.length < 3)
+        return false;
+    if (['.', '\u060C', '\u061F', '!', ' '].includes(sentence[0]))
+        return false;
+    return !/^(قبل|بعد)$/.test(sentence);
+    return !/^(برای|من|و|یا|آن|این|در|با|از|روی|قبل|بعد)$/.test(sentence);
```
gcampax, line 87: A comment with the English translation would be beneficial to future refactors of the code.

lib/replace_parameters.js
```diff
@@ -248,19 +238,17 @@ module.exports = class ParameterReplacer {
...
+                return ['string', 'tt:person_first_name'];
+            case 'tt:phone_number':
+                return ['string', 'tt:phone_number'];
+            case 'tt:hashtag':
+                return ['string', 'tt:word'];
+            case 'tt:path_name':
+                return ['string', 'tt:path_name'];
+
+            default:
+                return ['entity', entityType];
```
gcampax, line 251: Please preserve the existing format of switch statements, thanks.

(The existing format is **not** WebStorm's preferred format, which looks like what you're doing here)

lib/replace_parameters.js
```diff
@@ -506,6 +496,8 @@ module.exports = class ParameterReplacer {
                 }
                 if (slot.type.isLocation && !this._replaceLocations)
                     continue;
+                if ( (slot.type.isNumber || slot.type.type==='tt:phone_number') && !this._replaceNumbers)
+                    continue;
```
gcampax, line 500: Phone numbers are not numbers, they are PHONE_NUMBER_ tokens and are always preprocessed by the tokenizer.

If you want to handle PHONE_NUMBER_, you'll need to handle EMAIL_ADDRESS_, URL_ & PATH_NAME_ too.

lib/replace_parameters.js
```diff
@@ -583,7 +576,17 @@ module.exports = class ParameterReplacer {
...
+                        id: example.id,
+                        type: example.type,
+                        flags: example.flags || {},
+                        context: example.context,
+                        utterance: example.utterance,
+                        preprocessed: example.preprocessed,
+                        target_code: example.target_code
+                    };
+                    // console.error(e);
+                    // throw e;
```
gcampax, line 589: No. If there is an error, you're expected to crash. Otherwise the error goes unnoticed and we end up with broken datasets. The try-catch is only to add extra debugging info.

tool/augment.js
```diff
@@ -34,6 +34,11 @@ module.exports = {
             defaultValue: 'en-US',
             help: `BGP 47 locale tag of the language to generate (defaults to 'en-US', English)`
         });
+        parser.addArgument(['-pl', '--param_locale'], {
```
gcampax, line 37: In Genie, arguments use `-` not `_`.

tool/lib/file_parameter_provider.js
```diff
@@ -113,12 +113,12 @@ module.exports = class FileParameterProvider {
...
-        case 'entity':
-            return this._getEntities(valueListName);
-        default:
-            throw new TypeError(`Unexpected value list type ${valueListType}`);
+            case 'string':
+                return this._getStrings(valueListName);
+            case 'entity':
+                return this._getEntities(valueListName);
+            default:
+                throw new TypeError(`Unexpected value list type ${valueListType}`);
```
gcampax, line 121: Same here, please don't change the indentation of switch statements.

tool/requote.js
```diff
@@ -51,6 +52,10 @@ function findSpanType(program, begin_index, end_index) {
     let spanType;
     if (begin_index > 1 && program[begin_index-2] === 'location:') {
         spanType = 'LOCATION';
+    } else if (program[begin_index-2] === '==' && program[begin_index-3].endsWith('Number')){
+        spanType = 'NUMBER';
+    } else if (program[begin_index-2] === '==' && program[begin_index-3].includes('phone_number')){
+        spanType = 'PHONE_NUMBER';
```
gcampax, line 58: No. You cannot rely on type annotations, nor you can rely on having a param: token nearby.
If you want to detect a number, use a regexp that detects digits.
Note that unquoted numbers in thingtalk **do not** the `"` character. That is, the syntax is
```
now => @org.schema.Restaurant(), param:reviewCount >= 7 => notify;
```
rather than
```
now => @org.schema.Restaurant(), param:reviewCount >= " 7 " => notify;
```
(this is already existing syntax - if you invent new syntax you need to change thingtalk too)

As for phone_number/email_address/url, you should do exactly what the existing code does that for username/hashtag (that is, append `^^` and the entity identifier after the string literal)

Mehrad0711:
Thanks for the quick review! I am going through your comments...

lib/i18n/default.js
Mehrad0711: I mean in python we usually let base class return "NotImplementedError()". Here are real default is American English. I can change it though. Should I change "isGoodSentence" and "isGoodNumber" to true as well?

lib/dataset_augmenter.js
Mehrad0711: This is for when you wanna use a different language for parameters.
For translation task for example, I replace params with Farsi before translation so sentence is English but param is Farsi.

lib/dataset_augmenter.js
Mehrad0711: True that. I am only changing the requote and augment.
We will always have numbers as placeholders elsewhere. The numbers are replaced JUST before translation and then requoted RIGHT after translation.

lib/i18n/persian.js
Mehrad0711: yeah these two "\u060C|\u061F" are Arabic comma and question mark.
I will add "!" too.

lib/replace_parameters.js
Mehrad0711: lol  yeah  I did not intend to but WebStorm forced me on the indentations.
Is there a reason to not indent?

lib/replace_parameters.js
Mehrad0711: I am only doing so since numbers are problematic in translation.
As I mentioned above I only replace numbers before translation and requoted them after translation so other scripts will always see placeholders.

lib/replace_parameters.js
gcampax: It's how our style is defined. Style is pretty much arbitrary, but it needs to be consistent.

lib/i18n/default.js
gcampax: No the default is used for all languages for which we don't have language specific code. I.e. it will be used for Spanish, French, German, Japanese, etc.
So yeah you need a conservative default.

Mehrad0711:
Such pain to get the tests work...

lib/replace_parameters.js
gcampax: Yes I'm saying (here and the comment about lib/dataset_augmenter.js). you'll have problems in this script. Problems you don't see cause you have a limited dataset (restaurants) but will show up once you use it on a larger dataset (like Thingpedia).

tool/requote.js
```diff
@@ -19,6 +19,7 @@ const { maybeCreateReadStream, readAllLines } = require('./lib/argutils');
 const StreamUtils = require('../lib/stream-utils');
 
 const ENTITY_MATCH_REGEX = /^([A-Z].*)_[0-9]+$/;
+const NUMBER_MATCH_REGEX = /^([0-9|\u0660-\u0669]+)$/;
```
gcampax, line 22: Do you really need arabic numerals in the sentence? I don't think we want non-ASCII in ThingTalk syntax (outside of `"` marks). So you'd need to replace the numbers in the sentence using arabic numerals, and replace the code with regular numbers. Tricky, I suppose.

tool/requote.js
```diff
@@ -52,22 +53,23 @@ function findSpanType(program, begin_index, end_index) {
...
+                spanType = 'HASHTAG';
+                break;
+            case '^^tt:username':
+                spanType = 'USERNAME';
+                break;
+            case '^^tt:phone_number':
+                spanType = 'PHONE_NUMBER';
+                break;
+            default:
+                spanType = 'GENERIC_ENTITY_' + program[end_index+1].substring(2);
```
gcampax, line 86: Fix your webstorm ;)

tool/requote.js
```diff
@@ -21,6 +21,15 @@ const StreamUtils = require('../lib/stream-utils');
 const ENTITY_MATCH_REGEX = /^([A-Z].*)_[0-9]+$/;
 const NUMBER_MATCH_REGEX = /^([0-9|\u0660-\u0669]+)$/;
 
+function do_replace_numbers(token, requote_numbers) {
+    // check if token is:
+    // 1) an Arabic or English number
+    // 2) ignore digit 0 and 1 since it is sometimes used in program to represent
+    // singularities (e.g. a, an, one, ...) but is not present in the snetence
```
gcampax: "Small numbers" have been expanded to the range [-12, 12] recently.

tool/requote.js
```diff
@@ -49,12 +59,17 @@ function findSubstring(sequence, substring, spansBySentencePos) {
...
+function findSpanType(program, begin_index, end_index, requote_numbers) {
     let spanType;
     if (begin_index > 1 && program[begin_index-2] === 'location:') {
         spanType = 'LOCATION';
-    } else if (NUMBER_MATCH_REGEX.test(program[begin_index])){
-        spanType = 'NUMBER';
+
+    } else if (do_replace_numbers(program[begin_index], requote_numbers) && !(program[end_index+1].startsWith('^^'))){
+        // catch purely numeric postal code
+        if (program[begin_index - 3].endsWith('String'))
```
gcampax, line 68: NO. **Do not** assume that type annotations are present, and **do not** assume that there is a parameter nearby.

You know from the syntax if something is a number or a string. That's the whole point of `"` marks!

tool/requote.js
```diff
@@ -227,7 +246,7 @@ function getProgSpans(program, is_quoted) {
     } else {
         for (let i = 0; i < program.length; i++) {
             let token = program[i];
-            if (token === '"') {
+             if (token === '"') {
```
gcampax, line 248: Bogus indent

tool/requote.js
Mehrad0711: But the numbers which are not "0" or "1" are present in the sentence and requotable, right?

tool/requote.js
Mehrad0711: true, I will omit those characters since I'm using English digits for Farsi anyways.

lib/replace_parameters.js
Mehrad0711: Gotcha, the only answer I have now is replace_numbers only guaranteed to work for restaurant datasets :)) Can put a warning that "Be advised that it might not work on datasets containing measurements, etc."

gcampax: You can guess my answer ;)

PR #164, opened 2020-05-18
Changes happened before CIKM submission

rayslxu:


gcampax:
There's a lot of bug fix commits in this PR. Do you mind doing a round of `git rebase -i` and clean up the commit history for easier review?

rayslxu:
> There's a lot of bug fix commits in this PR. Do you mind doing a round of `git rebase -i` and clean up the commit history for easier review?

Sure.

rayslxu:
> There's a lot of bug fix commits in this PR. Do you mind doing a round of `git rebase -i` and clean up the commit history for easier review?

Merged some commits. There are two bug fixes not really associated to any commits, just something i realized during experiment. So I left them there. 
The last commit about complexity does a check on param distance, not sure if that's something we want in general.

tool/lib/bert-annotator.py
```diff
@@ -34,16 +34,32 @@ def template_query(cat, query_canonical, prefix, value='', suffix=''):
...
+        return [
+            f"show me a {query_canonical} {prefix} {value} {suffix} .".split(),
+            f"{question_start} is {prefix} {value} {suffix} .".split()
+        ]
     if cat == 'reverse_property':
-        return f"which {query_canonical} is a {prefix} {value} {suffix} ?".split()
+        return [
+            f"{question_start} is a {prefix} {value} {suffix} ?".split()
+        ]
     raise Exception('Invalid grammar category: ', cat)
```
gcampax, line 63: I feel like we should move this somewhere else, maybe some data file, to make it easy to extend in the future (for other languages, for other use cases, for other types of annotations)

languages/thingtalk/load-thingpedia.js
```diff
@@ -590,8 +590,8 @@ class ThingpediaLoader {
             'query',
             { p_name: Type.String },
             new Ast.Table.Filter(null, table, namefilter, table.schema),
-            [`\${p_name}`],
-            [`\${p_name}`],
+            [`\${p_name}`, ...canonical.map((c) => `\${p_name} ${c}`)],
+            [`\${p_name}`, ...canonical.map((c) => `\${p_name} ${c}`)],
```
gcampax, line 594: Strictly speaking, this is English only, we should use langPack and let other languages override this.

languages/thingtalk/ast_manip.js
```diff
@@ -1614,11 +1614,15 @@ function makeAggComputeArgMinMaxExpression(table, operation, field, list, result
...
+function makeArgMinMaxTable(table, pname, direction = 'desc', count) {
     if (hasUniqueFilter(table))
         return null;
     const sort = new Ast.Table.Sort(null, table, pname, direction, table.schema.clone());
-    return new Ast.Table.Index(null, sort, [new Ast.Value.Number(1)], sort.schema);
+    if (!count)
+        return sort;
+    if (count === 1)
+        return new Ast.Table.Index(null, sort, [new Ast.Value.Number(1)], sort.schema);
+    return new Ast.Table.Slice(null, sort, new Ast.Value.Number(1), new Ast.Value.Number(count), sort.schema);
```
gcampax: Slice([1:1]) should be autonormalized to Index([1]) if you want to simplify this code.

languages/thingtalk/en/computation.genie
```diff
@@ -242,6 +242,18 @@ compute_question = {
...
+        ('best' | 'top-rated' | 'top rated') table:with_filtered_table => {
+            if (!table.schema.out['aggregateRating.ratingValue'])
+                return null;
+            return C.makeArgMinMaxTable(table, 'aggregateRating.ratingValue', 'desc', 1);
+        };
+
+        ('best' | 'top') count:constant_Number table:with_filtered_table => {
+            if (!table.schema.out['aggregateRating.ratingValue'])
+                return null;
+            return C.makeArgMinMaxTable(table, 'aggregateRating.ratingValue', 'desc', count.value);
```
gcampax: What about big numbers? Those won't have .value, they're VarRefValue with name "__const_NUMBER"

languages/thingtalk/en/filters.genie
```diff
@@ -482,3 +482,8 @@ whose_npp_filter = {
     p:out_param_Any 'is' x:constant_Any => C.makeFilter(p, 'contains', x);
 }
 
+
+//FIXME: calculate threshold using max rating instead of a fixed number
+quality_filter = {
+    ('good' | 'great' | 'nice' | 'decent') => C.makeFilter(new Ast.Value.VarRef('aggregateRating.ratingValue'), '>=', new Ast.Value.Number(4));
```
gcampax, line 488: This is really weird to have in the domain-independent templates. It uses a schema.org specific parameter, and a schema.org specific value. I'd rather have webqa-process-schemaorg make up some dataset.tt automatically.

tool/webqa-process-schemaorg.js
```diff
@@ -344,6 +344,11 @@ const MANUAL_PROPERTY_CANONICAL_OVERRIDE = {
 
 };
 
+const MANUAL_TABLE_CANONICAL_OVERRIDE = {
+    'Restaurant': ['restaurant', 'diner', 'place', 'joint', 'eatery', 'canteen', 'cafeteria', 'cafe'],
+    'Hotel': ['hotel', 'resort', 'lodging', 'model', 'place']
```
gcampax, line 349: Wait, why do we have manual annotations for Hotel?

tool/webqa-update-canonicals.js
```diff
@@ -56,6 +56,12 @@ module.exports = {
             help: 'Skip the entire process.',
             defaultValue: false
         });
+        parser.addArgument('--no-neural', {
```
gcampax, line 59: This would be more understandable as `--only-heuristics`, or even better `--algorithm={heuristic,neural}`

tool/webqa-process-schemaorg.js
rayslxu: This is only enabled when we enable manual annotation. I used this to try out the performance of hotels.

languages/thingtalk/load-thingpedia.js
rayslxu: where is langPack?

tool/webqa-process-schemaorg.js
gcampax: Don't we enable manual annotation in hotels to get the manual annotation of restaurant fields?

languages/thingtalk/load-thingpedia.js
gcampax: `this._langPack` in load-thingpedia is a reference to a file in lib/i18n
Anyway I didn't mean this as a real suggestion for this PR, more of mentioning it.
(Now that I think about it, you could also get the same effect as a standard Genie template, using `constant_name` and `base_noun_phrase`. Anyway not worth the effort.)

tool/webqa-process-schemaorg.js
rayslxu: yes, I had a separate genie to run that. I think as the toolkit, if you say manual, you get full manual.

tool/webqa-update-canonicals.js
rayslxu: updated with c3aa6e0

tool/webqa-process-schemaorg.js
gcampax: That makes sense. Although you seriously need to document this, so many commands, even I have no idea what does what...

languages/thingtalk/en/filters.genie
rayslxu: the problem is that we do not have a way to apply "good restaurant" in dataset.tt to generate "good chinese restaurant", we need to update how we use primitive templates for that ...

languages/thingtalk/en/computation.genie
rayslxu: how do i get the actual value for that?

tool/webqa-process-schemaorg.js
rayslxu: yeah, agree. i will probably do that after emnlp

languages/thingtalk/en/filters.genie
gcampax: We really need domain-specific templates with the full (or close to full) power of Genie templates... But that's a matter for another day.

languages/thingtalk/en/computation.genie
gcampax: You don't. You do everything with Ast.Value, instead of taking the numeric value and then wrapping it again into a NumberValue

rayslxu: thanks. fixed in a467169

languages/thingtalk/ast_manip.js
rayslxu: updated in a467169

rayslxu:
> Quite a few comments throughout, and a bug around "top rated restaurants". But overall pretty good.

I think I addressed all the issues except for those more long term ones. 
I also have two new commits
f7171f3 where i splitted the manual annotations out from webqa-process-schemaorg, since some of them are shared across scripts
2e85d90 we skip building string set for non-filterable params - this fix the time-consuming tokenizing issue for description field

PR #175, opened 2020-06-17
Add new annotations

gcampax:
This PR implements new useful annotations that are not dialogue specific.


gcampax:
Ugh unfortunately a7cef46 breaks requoting, and there is no good way to fix it. Converting back to draft while I think about this...

gcampax:
Yes we'll update the documentation after this round of cleanup.

PR #178, opened 2020-06-21
Rewrite the tokenizer, kill almond-tokenizer

gcampax:
This is the beginning of the cleanup I promised in chat (and I had meant to do for forever).

It replaces almond-tokenizer will a regexp-based preprocessor, implemented entirely in JavaScript. The preprocessor does the usual stuff: dates, times, phone numbers, emails, filenames, etc.

It does less than what CoreNLP does in terms of keeping special symbols together (which is helpful when integrating with BERT down the line). There is no statistical tokenization of Chinese, all tokens are single characters. This is ok because Chinese uses XLM-R which uses SentencePiece, so we just need to remove extra spaces before we feed to the network.

Important differences for English (please review carefully):
- hyphens are never split (no clever heuristics)
- fractions are not recognized, and / are always split unless recognized as a filename
- underscores in the middle of a word cause a word split
- assimilations (cannot, don't) are not split
- measurements without space ("5gb") are split (unless recognized as a time or postcode)
- no legacy encoding of parenthesis (-lrb- / -rrb-), which means those characters can now appear in the dataset
- Unicode compatibility normalization is applied - this splits certain exotic characters into ASCII forms, for example ½ becomes "1/2" which becomes "1 / 2" (this appears often in the recipe dataset)

Minor differences and things that are changed for the better:
- abbreviations might be slightly different than CoreNLP (we'll need to check with tt:short_free_text, tt:long_free_text and similar datasets to ensure no change)
- parenthesis, dashes and quotes are not transformed
- no Americanization (colour -> color, € -> $)
- ordinals in words are now recognized (subject to the usual small number rules)

New feature:
- The tokenizer produces a "normalized" sentence representation which cuts a lot of the tokenization noise and normalizes numbers/dates/times. It is designed to capture all information in the sentence and is idempotent (you tokenize it again you get the same string out). We could experiment with feeding that to the neural network in place of the TIME/DATE tokens.

Things I'm unsure about:
- durations (not implemented) - should we treat these like regular measurements?
- currencies (implemented as before) - same question, does it make sense to have a fixed list of 5/6 currency codes?
- handling of year/month without a day - in almond-tokenizer, it means today, but in that month, which is not very intuitive
- colloquial times for English (and Italian) are quite limited - basically only "o'clock" is recognized; I'm not sure how much we need to worry about those, it depends a lot of what the speech to text does

Languages implemented:
- English, because of course
- Italian, to see how hard it is to add a new language
- Chinese, to see how well we do with non-spacing languages


gcampax:
Btw, the only language specific parts here are numbers, ordinals, dates and times.
For dates and times, the [CLDR project](https://github.com/unicode-cldr/cldr-json) has basically already done the work of catalogizing how each language expresses them, and it's only a matter of converting from a generator of dates to a parser of dates. For numbers and ordinals, you can find guides on the Internet that explain how to say numbers in all numbers. See e.g. for [French](https://www.thefrenchexperiment.com/learn-french/numbers) and [German](https://www.fluentin3months.com/german-numbers/).
Which is to say, a sufficiently motivated person could go ahead and write all the other 7 languages that we care about.

lib/i18n/tokenizer/chinese.js
```diff
@@ -0,0 +1,303 @@
...
+//
+// Author: Giovanni Campagna <gcampagn@cs.stanford.edu>
+//
+// See COPYING for details
+"use strict";
+
+const BaseTokenizer = require('./base');
+const { makeToken } = require('./helpers');
+
+// 我觉得这比英语太好了
```
rayslxu, line 15: // 我觉得这比英语好太多了

gcampax: I take it back. English would not betray me in a comment.

gcampax:
> 
> How is measurement handled? I couldn't locate the part for it.

Regular measurements are not handled by almond-tokenizer, and not handled by the new code either: the number becomes a NUMBER and the neural network will copy the NUMBER and predict a "unit:" token.
(MEASURE_ tokens do exist, but they exist only while synthesizing dialogues and go away when we prepare the data for the neural network)

DURATIONS are not handled, as I mentioned in the cover letter, and I'm not sure they should be.

lib/i18n/tokenizer/chinese.js
```diff
@@ -0,0 +1,303 @@
...
+            const value = this._parseDecimalNumber(lexer.text);
+            if (Math.floor(value) === value && value <= 12 && value >= -12)
+                return makeToken(lexer.index, lexer.text, String(value));
+            else
+                return makeToken(lexer.index, lexer.text, String(value), 'NUMBER', value);
+        });
+
+        // currencies
+        this._lexer.addRule(/{DECIMAL_NUMBER}(?:[美加澳欧]?元|英镑|日圆)/, (lexer) => {
+            let unitlength = /[美加澳欧]元|英镑|日圆/.test(lexer.text) ? 2 : 1;
```
rayslxu, line 132: "日圆" is more Japanese, we use "日元" normally.

lib/i18n/tokenizer/chinese.js
```diff
@@ -0,0 +1,303 @@
...
+        } else {
+            return { year, month, day, hour: 0, minute: 0, second: 0, timezone: undefined };
+        }
+    }
+
+    _initDates() {
+        // init ISO date recognition
+        super._initDates();
+
+        this._addDefinition('DAY_NAME', /星期[一二三四五六日天]/);
```
rayslxu, line 275: `/(星期|周|礼拜)[一二三四五六日天]/` we have 3 different ways to say "week"

lib/i18n/tokenizer/chinese.js
```diff
@@ -0,0 +1,303 @@
...
+
+        this._addDefinition('DAY', /(?:三十一?|[一二]十[一二三四五六七八九]?|[一二三四五六七八九]|3[01]|[12][0-9]|[1-9])号/);
```
rayslxu, line 277: I'm surprised you pick up “号", it's more colloquial, but it's almost never used in written language, we use "日“ instead.

gcampax:
> I'm not sure why we would treat duration/currency differently, it seems unnecessary.
> I feel like month/year without a date should be considered as a duration instead - it was one of the road block for me in movie/music domain - 2020 movies should be movies with publish date in duration 2020.

Your use case would be a date range (a pair of dates) not a duration. A duration is just "one month".
FWIW, if you pick the month/year to be the first day of the month, you can train the network to do "param >= DATE_0 && param <= DATE_0 + 1 unit:mon" but that's not easy to synthesize. So maybe we need a special type too.

As for Currency, if we get rid of currency preprocessing we'll need new syntax in ThingTalk, and we'll need to update our datasets (Uber, Bitcoin, Stock price, Steam) to replace "CURRENCY_0" with "$ NUMBER_0" or "NUMBER_0 dollars". But it might be worthwhile anyway.

PR #183, opened 2020-06-23
Two small fixes for ReplaceParameters

gcampax:


lib/replace_parameters.js
```diff
@@ -359,7 +359,7 @@ module.exports = class ParameterReplacer {
                 operator = '==';
         }
 
-        let attempts = 10000;
+        let attempts = Math.min(10000, valueList.length);
         while (attempts > 0) {
```
Mehrad0711, line 363: Since we are doing sampling with replacement this can be problematic for smaller parameter datasets (especially when you use --clean-parameters flag). Can we make this optional?   Btw, you can have number of maximum attempts as a flag and adjust according to your parameter datasets size. (I have that in my PR)

gcampax: You can make it the square of the dataset size if you want, and then by birthday paradox you know with high probability you will sample everything once.

Mehrad0711: Making it square results in similar performance cause our param datasets usually have at least 100 samples. Is there a fundamental reason you want this number to be caped? In most cases, if it fails to replace after 10000 attempts mean that the param dataset is problematic which we have to fix.

gcampax: Exactly. Except it keeps trying 10000 times for every sentence, which makes it very slow. Maybe what we should do is notice that a certain slot failed to replace, and not even try it again.

Mehrad0711: Yeah, my concern is mainly for when --clean-parameters is used since that actually needs several attempts. Let's have this change (maybe use "5*valueList.length" instead) and cap attempts but only when --clean-parameters is false?

PR #186, opened 2020-06-26
Autoqa code reorganization

rayslxu:
Remove deprecated webqa name, reorganize scripts around (move schema.org specific scripts under autoqa folder).


tool/autoqa/normalize-data.js
```diff
@@ -526,7 +526,7 @@ class Normalizer {
 
 module.exports = {
     initArgparse(subparsers) {
-        const parser = subparsers.addParser('webqa-normalize-data', {
+        const parser = subparsers.addParser('autoqa-normalize-data', {
```
gcampax, line 529: This one is schema.org specific right? Unlike the other ones. Perhaps this should be reflected in the command name.

rayslxu:
> It's just a rename, right?

Yeah, mostly rename with some minor changes. All autoqa code is in fact schema.org specific, i can call it schema2qa maybe?

PR #190, opened 2020-07-01
Clean-up lib/i18n

gcampax:
Make everything into classes, and remove duplicated and identical code.

@Mehrad0711 please look carefully at this. I recommend using the "ignore white space" option of the diff (`-w` in the command-line). Also check: some isGoodNumber implementations include Persian numbers, even for languages that use the latin alphabet.
Japanese also might need work, because it's an ideographic language but it's treated like an alphabetic one.


Mehrad0711:
looks good. Need to update the isGoodNumber for some of the languages. Chinese numbers should be added to the Chinese package. 
Also, Spanish and Japanese isGoodSentence can be improved. Shall I fix those in this PR?

gcampax:
Please make a separate PR. This one is already going to cause conflicts.

I'm still unclear on the design/purpose of isGoodNumber by the way - shouldn't numbers always be in ASCII digits? The tokenizer normalizes the numbers in words (incl. Chinese characters) to numbers in ASCII digits.

PR #191, opened 2020-07-02
Documentation and tests

gcampax:
I want to refactor again how dialogues are generated, but to do so, I want to have a pipeline to test my changes (without going through our AWS infrastructure and training real models). The best way is to bring in the Makefiles we use for homeworks and tutorials, which this PR does.


rayslxu:
Should we add documentations for the starter code for schema.org and multiwoz as well?

gcampax:
We should, yes. I'll let you document the starter code for schema.org, and I'll deal with the multiwoz stuff when it becomes more stable.

PR #194, opened 2020-07-08
Dialogues, again

gcampax:
This PR contains the big rewrite of the state machine that I talked about.

It also contains all the fixes necessary to get MultiWOZ eval set to behave on the master branch, with ThingTalk from https://github.com/stanford-oval/thingtalk/pull/217

I'm going to test this PR with a proper dataset MultiWOZ and then unmark it as draft.
It is on top of the "next" branch (which has the big import of almond-dialog-agent/thingengine-core) but I think once "next' is in shape enough to run autoqa and multiwoz, we should merge "next" to "master" and continue there.


gcampax:
Btw I discovered a bug in the current dialogues: when parameter passing from hotels/restaurans/attractions to taxi, we fail to identify the taxi location from the context, because the value in the context and the value in the taxi program have different types (one is an entity and the other is a string). We then assume the program is misannotated, and tell the network to predict the string directly, which obviously won't work because the network doesn't see the string in the context...
This probably causes us to miss all parameter passing.

gcampax:
I tested this by generating a multiwoz dataset and it seems to go through, so I think you can start looking at it.

PR #195, opened 2020-07-09
autoqa updates

rayslxu:
Reorganize autoqa scripts, add scripts for sgd dataset, as well as some wikidata scripts.


tool/lib/base-canonical-generator.js
```diff
@@ -0,0 +1,70 @@
...
+//
+// Author: Silei Xu <silei@cs.stanford.edu>
+//
+// See COPYING for details
+"use strict";
+
+const { clean } = require('../../lib/utils');
+const { pluralize, posTag } = require('../../lib/i18n/american-english');
+
+function genBaseCanonical(canonical, name, ptype) {
```
gcampax: Is this something that we want to run automatically for all Thingpedia devices if one does not specify `#_[canonical]` when uploading?
(Currently, we only run `clean`)

tool/sgd/utils.js
```diff
@@ -0,0 +1,23 @@
...
+//
+// See COPYING for details
+"use strict";
+
+module.exports = {
+    cleanEnumValue(v) {
+        // replace dash with space
+        v = v.replace(/-/g, ' ');
+        // camelcase the value
+        v = v.replace(/(?:^|\s+|-)[A-Za-z]/g, (letter) => letter.trim().toUpperCase());
```
gcampax: camelCase? Not snake_case ?

tool/autoqa/sgd/normalize-data.js
```diff
@@ -136,13 +89,13 @@ class Normalizer {
...
-                return parseDuration(value);
+                return ThingTalk.Units.transformToBaseUnit(parseFloat(value), 'min');
             } else if (expectedType.type === 'tt:Measure') {
-                return parseMeasure(value);
+                throw new Error(`No measurement type in sgd dataset`);
             } else if (expectedType.type.startsWith('tt:Enum(')) {
                 const enumerands = expectedType.type.substring('tt:Enum('.length, expectedType.type.length - 1).split(/,/g);
                 value = cleanEnumValue(value);
-                if (value === undefined)
+                if (value === undefined || value === 'Dontcare')
```
gcampax, line 102: Do we need to handle Dontcare values correctly (map to a BooleanExpression.DontCare)?

rayslxu:
>     * why does "make-string-dataset" depend on `--dataset` ? The format is normalized at that point, and the thingpedia.tt file has all the information.

It uses some information in the corresponding manual annotations. Also, for sgd, each parameter is shared among multiple intents in one service, so there is some small difference.

gcampax:
> It uses some information in the corresponding manual annotations. Also, for sgd, each parameter is shared among multiple intents in one service, so there is some small difference.

But there can only be one database query per entity, so all the different query intents get merged, and actions don't count, right?

tool/sgd/utils.js
rayslxu: That's what we did for schema.org...

tool/lib/base-canonical-generator.js
rayslxu: I think so - I suppose I should move it to `lib` as you suggested last time?

tool/autoqa/sgd/normalize-data.js
rayslxu: Yes, but in step we are generating normalized data, so this should not matter? But yeah, we need to take care of this when transforming their dataset to our thingtalk annotation

rayslxu:
> Yeah looks good, a couple comments.
> Also, a couple high-level things I noticed:
> 
>     * why does "make-string-dataset" depend on `--dataset` ? The format is normalized at that point, and the thingpedia.tt file has all the information.
> 
>     * do we want to unify all the "process-schema" scripts into one with a `--dataset` option?
> 
>     * do we want a `--dataset custom` option to "auto-annotate" that does not include any manual annotation and can run on arbitrary Thingpedia file? (that would probably be the same as `--dataset multiwoz`, given that multiwoz .tt files were written by hand)



> > It uses some information in the corresponding manual annotations. Also, for sgd, each parameter is shared among multiple intents in one service, so there is some small difference.
> 
> But there can only be one database query per entity, so all the different query intents get merged, and actions don't count, right?

Actually not, their actions are different from multiwoz. e.g, the signature of their `FindRestaurant` and `ReserveRestaurant` is pretty much the same - you don't append an action after the query, you just replace the query with the action.

gcampax:
Yeah that's not going to work. You need to fit into the state machine, or you'll get nothing out. MultiWOZ doesn't "naturally" come with our design, it was made to fit into it. SGD will need the same.

gcampax:
(Incidentally, this is why I insist you look at the dialogue templates. The more you wait, the more work you waste...)

PR #206, opened 2020-07-11
Priority controls for the dialogue agent

gcampax:
Letting the dialogue agent behave randomly is not good, especially with a policy that is completely exploratory.


gcampax:
You still get a little bit of randomness in terms of the specific wording, when it comes to sentence chunks in `( 'a' | 'b' )`, but yeah, not having randomness is somewhat intentional. I think for now I want to err on predictability rather than randomness, especially as the priority are not that well tuned and the agent is still rough. We can perhaps revisit later.

PR #210, opened 2020-07-15
New round of templates for Thingpedia

gcampax:
This is part 1 of a series of PR I will be doing in the next few days/weeks, implementing missing functionality that is required to access Thingpedia using the state-machine-based dialogue agent.

This PR includes support for queries that return a single result (such as weather), queries that return a random result, and queries with input parameters.
I'm sending this work in small PRs so we can merge it quickly because SGD and similar will need this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,10 +21,6 @@ po/*.json
 *.tt.json
 
 workdir*
-dataset.tt
-thingpedia.json
-entities.json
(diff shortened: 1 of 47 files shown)


gcampax:
Note: tests won't pass because I haven't released a version of ThingTalk with the entity definition statements. Also tests need updating probably because the templates changed.

PR #213, opened 2020-07-17
force manual annotations to use same entityRetriever as in eval

RyanOthnielKearns:


gcampax:
> During syntactic normalization, setting `allocateEntities: true` allows the manual annotator to ignore the sentence and resolve `entities` straight from TT allocation, ignoring the sentence. This is different from the `entityRetriever` used in the evaluation step, which must look at the sentence for entity resolution. Changing the `entityRetriever` during manual annotation thus provides a sanity check and prevents the user from introducing an entity lookup error during evaluation.

This blurb is very useful! You should put it in the commit message.

RyanOthnielKearns:
> I think this is correct, although I hope you tested this because we don't yet have automatic tests of the annotation tools.

Can confirm I've tested this locally; manual annotations with improper entity resolution will fail just as they do in eval, and provide the user with a `TT: ` prompt as if they'd committed a syntax error.

gcampax:
Yeah go for it then.

PR #214, opened 2020-07-17
Auto annotate multiwoz

rayslxu:
- allow bert to run when string set is not available
- run bert on boolean annotations

This pr only make autoqa run on multiwoz, it does not improve the annotation strategy.

### DIFF
--- a/tool/autoqa/lib/bert-canonical-annotator.py
+++ b/tool/autoqa/lib/bert-canonical-annotator.py
@@ -137,6 +137,8 @@ class BertLM:
             for arg in queries[query]['args']:
                 if 'path' in queries[query]['args'][arg]:
                     self.values[query][arg] = self.load_values(queries[query]['args'][arg]['path'])
+                elif 'values' in queries[query]['args'][arg]:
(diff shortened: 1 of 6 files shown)


tool/autoqa/lib/bert-canonical-annotator.py
```diff
@@ -46,20 +56,44 @@ def template_query(cat, query_canonical='', prefix='', value='', suffix=''):
             f"show me a {query_canonical} with {prefix} {value} {suffix} .".split(),
             f"{question_start} has {prefix} {value} {suffix} ?".split()
         ]
+    if cat == 'property_true' or cat == 'property_false':
+        return [
+            f"show me a {query_canonical} with {prefix} .".split(),
+            f"{question_start} has {prefix} ?".split()
```
gcampax, line 62: I would add a comment here noting that property_false also uses "with" instead of "without" (ie, we're trying to form "show me a restaurant with no parking" as opposed to "show me a restaurant without parking"), cause that tripped me just now.

PR #216, opened 2020-07-18
Add simple action rejection templates

RyanOthnielKearns:
Currently, `sys_confirm_action` states are only followed by the user accepting the action. This PR adds two additional transitions:
1. Action cancellations, of the utterance form "No, thank you." or "Actually, not yet.". These are annotated as `cancel` states.
2. Action modifications, of the utterance form "Actually, could you do [X] instead?". These are annotated with the new action.

### DIFF
--- a/languages/thingtalk/dialogue_acts/action-confirm.js
+++ b/languages/thingtalk/dialogue_acts/action-confirm.js
@@ -22,6 +22,7 @@
 const assert = require('assert');
 
 const ThingTalk = require('thingtalk');
+const Ast = ThingTalk.Ast;
 const Type = ThingTalk.Type;
 
 const C = require('../ast_manip');
@@ -30,6 +31,8 @@ const {
     makeAgentReply,
     makeSimpleState,
     sortByName,
+    setOrAddInvocationParam,
+    replaceAction,
 } = require('../state_manip');
 
 
@@ -68,7 +71,38 @@ function actionConfirmAcceptPhrase(ctx) {
     return clone.state;
 }
 
+function actionConfirmRejectPhrase(ctx) {
+    // remove last action in history so it's not carried over into the cancel state
+    return makeSimpleState(ctx, 'cancel', null, true);
+}
+
+// function actionConfirmChangeParam(ctx, answer) {
+//     // console.log(ctx);
+//     // console.log(answer);
+//     const schema = ctx.nextFunctionSchema;
+//     const questions = ctx.dialogueActParam || [];
+//     if (answer instanceof Ast.Value) {
+//         if (questions.length !== 1)
+//             return null;
+//         answer = new Ast.InputParam(null, questions[0], answer);
+//     }
+//     const arg = schema.getArgument(answer.name);
+//     if (!arg || !arg.is_input || !arg.type.equals(answer.value.getType()))
+//         return null;
+
+//     // console.log(ctx.state.history[0]);
+//     const action = C.getInvocation(ctx.state.history[0]);
+//     // console.log(action);
+//     if (!action)
+//         return null;
+//     const clone = action.clone();
+//     setOrAddInvocationParam(clone, answer.name, answer.value);
+//     return replaceAction(ctx, 'execute', clone, 'accepted');
+// }
+
 module.exports = {
     makeActionConfirmationPhrase,
-    actionConfirmAcceptPhrase
+    actionConfirmAcceptPhrase,
+    actionConfirmRejectPhrase
+    // actionConfirmChangeParam
 };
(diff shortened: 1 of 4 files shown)


languages/thingtalk/dialogue_acts/action-confirm.js
```diff
@@ -68,7 +72,37 @@ function actionConfirmAcceptPhrase(ctx) {
     return clone.state;
 }
 
+function actionConfirmRejectPhrase(ctx) {
+    return new Ast.DialogueState(null, POLICY_NAME, 'cancel', null, []);
```
gcampax, line 76: This should use `makeSimpleState` instead.

languages/thingtalk/dialogue_acts/action-confirm.js
```diff
@@ -68,7 +72,37 @@ function actionConfirmAcceptPhrase(ctx) {
...
+
+function actionConfirmChangeParam(ctx, answer) {
+    const questions = ctx.dialogueActParam || [];
+    if (answer instanceof Ast.Value) {
+        if (questions.length !== 1)
+            return null;
+        answer = new Ast.InputParam(null, questions[0], answer);
+    }
+
+    const action = C.getInvocation(ctx.state.history[0]);
```
gcampax, line 87: You don't want to look at the first result in the history, you want to look at the next one (`ctx.next`).

languages/thingtalk/dialogue_acts/action-confirm.js
```diff
@@ -68,7 +72,37 @@ function actionConfirmAcceptPhrase(ctx) {
...
+    if (!action) return null;
+
+    // don't continue with queries
+    if (action.schema._functionType === "query") return null;
```
gcampax, line 91: I'm not sure why this check? We want to say no to confirmable queries too.
Also, `_` indicates a private property. Don't use those outside of the specific class.

languages/thingtalk/dialogue_acts/action-confirm.js
```diff
@@ -68,7 +72,37 @@ function actionConfirmAcceptPhrase(ctx) {
     return clone.state;
 }
 
+function actionConfirmRejectPhrase(ctx) {
+    return new Ast.DialogueState(null, POLICY_NAME, 'cancel', null, []);
+}
+
+function actionConfirmChangeParam(ctx, answer) {
+    const questions = ctx.dialogueActParam || [];
```
gcampax, line 80: There won't be any questions at this template - it comes out of the agent asking for a confirmation, not asking for a parameter.

languages/thingtalk/dialogue_acts/action-confirm.js
```diff
@@ -68,7 +72,37 @@ function actionConfirmAcceptPhrase(ctx) {
...
+    const action = C.getInvocation(ctx.state.history[0]);
+    if (!action) return null;
+
+    // don't continue with queries
+    if (action.schema._functionType === "query") return null;
+
+    // don't accept in params that don't apply to this specific action
+    const allowedParams = new Set(Object.keys(ctx.nextFunctionSchema._inReq));
+    allowedParams.add(Object.keys(ctx.nextFunctionSchema._inOpt));
+    if (!allowedParams.has(answer.name)) return null;
```
gcampax, line 96: It's easier to write this as
```
const arg = ctx.nextFunctionSchema.getArgument(answer.name);
if (!arg || !arg.is_input)
```
You also need to check that the type is correct.

languages/thingtalk/dialogue_acts/action-confirm.js
RyanOthnielKearns: Good to know about `_`. Removing the check doesn't seem to cause any issue with generation. Though, it seems a misnomer to assign `sys_confirm_action` to a confirmable query.

gcampax: It does, but having sys_confirm_action and sys_confirm_query would be worse, and also confirmable queries are only a theoretical possibility for now, because queries don't have side effects so they're always safe to run.

RyanOthnielKearns: Ah, makes sense!

RyanOthnielKearns:
> Did you test this?
> A good way to test templates is to comment out all other templates and run a round of generation. You can force this to fire on multiwoz if you set `#[confirm=true]` on the multiwoz actions, if you haven't gotten to SGD yet.

Yes, this has been tested on Mario's Twitter experiment and generates fine. Generation also looks good on MultiWOZ, though `make datadir` fails occasionally there for a previous issue (i.e., something wrong also in `master`). SGD doesn't yet have a `schema.tt` so I have yet to test it there.

PR #220, opened 2020-07-20
Small changes for requote and augment scripts

Mehrad0711:


tool/requote.js
```diff
@@ -457,44 +459,65 @@ module.exports = {
...
-                            callback(null);
-                        else
+                        ex.is_ok = false;
+                        if (args.skip_errors) {
+                            this.push(ex);
+                        } else {
                             callback(e);
+                        }
                     }
+                    callback();
```
gcampax, line 508: You're calling callback twice if skip_errors is not set - this will cause a crash.

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -202,7 +202,7 @@ module.exports = class ParameterReplacer {
         this._blowUpSynthetic = _default(options.syntheticExpandFactor, 5);
         this._blowUpNoQuote = _default(options.noQuoteExpandFactor, 10);
         this._blowUpParaphrasing = _default(options.paraphrasingExpandFactor, 30);
-        this._blowUpAugmented = Math.floor(this._blowUpParaphrasing/2);
+        this._blowUpAugmented = Math.max(1, Math.floor(this._blowUpParaphrasing/2));
```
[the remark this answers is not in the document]
Mehrad0711: yeah, I was running it on some prehistoric datasets and saw output size is smaller than the input. Debugging led me to this that blowUpAugmented becomes 0 when all blowUpParaphrasing is 1.

lib/dataset-tools/simple-splitter.js
```diff
@@ -0,0 +1,57 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author: Mehrad Moradshahi <mehrad@cs.stanford.edu>
+"use strict";
+
+const Stream = require('stream');
+
+
+module.exports = class SimpleDatasetSplitter extends Stream.Writable {
```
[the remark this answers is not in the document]
Mehrad0711: addressed.

lib/dataset-tools/conditional-splitter.js
```diff
@@ -0,0 +1,61 @@
...
+"use strict";
+
+const Stream = require('stream');
+
+/**
+ This class provides an interface to conditionally control the flow of input stream
+ to two writable streams based on 'is_ok' property of each input
+ */
+
+module.exports = class ConditionalDatasetSplitter extends Stream.Writable {
```
gcampax, line 29: jsdoc has a bug where it won't attach the documentation to the class if the class is directly assigned to `module.exports`
You need to do:
```
class Foo { ... }
module.exports = Foo
```

Mehrad0711: fixed.

PR #235, opened 2020-07-21
Fix bert annotation and improve speed

rayslxu:
I introduced a bug which makes both bert and bart to run on all values instead of sampled examples.
This PR makes the test run in a not too crazy long time at least. 
Future improvement on Bart will be in a separate PR

### DIFF
--- a/tool/autoqa/lib/bert-canonical-annotator.py
+++ b/tool/autoqa/lib/bert-canonical-annotator.py
@@ -165,15 +165,21 @@ class BertLM:
         self.pruning_threshold = pruning_threshold
         self.queries = queries
         self.canonicals = {}  # canonical of queries
-        self.values = {}  # values of arguments
+        self.values = {}  # all the values of each argument
+        self.example_values = {}  # example values (constants.tsv) of each argument
         for query in queries:
             self.canonicals[query] = queries[query]['canonical']
             self.values[query] = {}
+            self.example_values[query] = {}
             for arg in queries[query]['args']:
+                if 'values' in queries[query]['args'][arg]:
+                    self.example_values[query][arg] = self.queries[query]['args'][arg]['values']
+                else:
+                    self.example_values[query][arg] = []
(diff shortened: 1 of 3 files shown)


tool/autoqa/lib/bert-canonical-annotator.py
```diff
@@ -151,7 +153,7 @@ def __init__(self, queries, mask, k_synonyms, k_adjectives, pruning_threshold, m
         self.tokenizer = BertTokenizer.from_pretrained(model_name_or_path)
 
         # Load pre-trained model (weights)
-        self.model = BertForMaskedLM.from_pretrained(model_name_or_path)
+        self.model = BertForMaskedLM.from_pretrained(model_name_or_path).to('cuda')
```
gcampax, line 156: .to(device) ?

tool/autoqa/lib/bert-canonical-annotator.py
```diff
@@ -165,15 +167,21 @@ def __init__(self, queries, mask, k_synonyms, k_adjectives, pruning_threshold, m
...
+                    self.example_values[query][arg] = self.queries[query]['args'][arg]['values']
+                else:
+                    self.example_values[query][arg] = []
                 if 'path' in queries[query]['args'][arg]:
                     self.values[query][arg] = self.load_values(queries[query]['args'][arg]['path'])
                 elif 'values' in queries[query]['args'][arg]:
-                    self.values[query][arg] = self.queries[query]['args'][arg]['values']
+                    self.values[query][arg] = self.example_values[query][arg]
                 else:
                     self.values[query][arg] = []
```
gcampax, line 186: So if I understand this bit correctly, queries both have "path" and "values" and you want "values" to win instead of "path" when setting "example_values", which you use later.
Where do you use "values"?

rayslxu: `values` is used to identify adjectives - where we find matches among all values.

PR #246, opened 2020-07-27
Two small bug fixes

Mehrad0711:


lib/dataset-tools/flags.js
```diff
@@ -19,6 +19,8 @@
 // Author: Giovanni Campagna <gcampagn@cs.stanford.edu>
 "use strict";
 
+const flagsMap = {'R':'replaced', 'P':'augmented', 'C':'contextual', 'S':'synthetic', 'E':'eval'};
```
gcampax, line 22: Add a comment that the order of entries is significant, because the order of flags is significant when parsing them.

(The order matters because there are multiwoz dialogues with ID that starts with "SNG" - but they should not get the "S" flag so we prefix the ID with "E", eval)

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -260,7 +260,7 @@ module.exports = class ParameterReplacer {
             slot.tag === 'program.principal'))
             return ['string', 'tt:person_first_name'];
 
-        if (!slot.type.isEntity && !slot.type.isString && !slot.type.isLocation && !slot.type.isNumber)
+        if (!slot.type.isEntity && !slot.type.isString && !slot.type.isLocation && !slot.type.isNumber && !slot.type.isMeasure)
```
gcampax, line 263: Do we handle measurements correctly though?

Mehrad0711: They are currently treated the way numbers are treated (assuming measurement units are included in the original sentence)

gcampax: Ok

PR #247, opened 2020-07-27
Add script to check how well synthetic set covers evaluation set

rayslxu:
Examples: 
https://docs.google.com/spreadsheets/d/1TuYwpAEwd0-t7OwGhyG4oXVGwWzNI1mnuEku6tHGcNk/edit?usp=sharing


tool/compute-synthetic-coverage.js
```diff
@@ -0,0 +1,106 @@
...
+        parser.addArgument('--synthetic-set', {
+            required: true,
+            type: fs.createReadStream,
+            help: 'Path to the file containing the synthetic data (in TSV format: id, utterance, thingtalk)'
+        });
+        parser.addArgument('--evaluation-set', {
+            required: true,
+            type: fs.createReadStream,
+            help: `Path to the file containing the evaluation data (in TSV format: id, utterance, thingtalk)`
+        });
```
gcampax, line 48: `--contextual`? Or nah?

tool/compute-synthetic-coverage.js
```diff
@@ -0,0 +1,106 @@
...
+    },
+
+    async execute(args) {
+        let syntheticSize = 0;
+        const syntheticPrograms = {}; // counter of unique programs in synthetic set
```
gcampax, line 53: You'll probably want to use a Map instead of a plain object, it's a lot faster when you have thousands of keys.

tool/compute-synthetic-coverage.js
```diff
@@ -0,0 +1,106 @@
...
+        args.synthetic_set.setEncoding('utf8');
+        const synthetic = args.synthetic_set.pipe(csvparse({ relax: true, delimiter: '\t' }));
```
gcampax, line 55: DatasetParser???

tool/compute-synthetic-coverage.js
```diff
@@ -0,0 +1,106 @@
+// -*- mode: js; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+//
+// This file is part of Genie
+//
+// Copyright 2019-2020 The Board of Trustees of the Leland Stanford Junior University
```
gcampax, line 5: Copyright year.

rayslxu:
> I assume you coordinated with Ryan & Lucas? We had basically the same analysis in our paper to the Conference-That-Cannot-Be-Named, and we had a fairly sophisticated python script for this.

No! I was not aware of that. 
which section is the analysis in? and where is the python code?

gcampax:
The analysis is section 5.4. The scripts are in genie-workdirs/dlgthingtalk, but I haven't touched in a while so you should check.

tool/compute-synthetic-coverage.js
```diff
@@ -55,7 +55,8 @@ module.exports = {
         const synthetic = args.synthetic_set.pipe(csvparse({ relax: true, delimiter: '\t' }));
         synthetic.on('data', (line) => {
             const requoted = Array.from(requoteProgram(line[2])).join(' ');
-            syntheticPrograms[requoted] = (syntheticPrograms[requoted] || 0) + 1;
+            const requotedNoType = requoted.replace(/(param:[a-zA-Z_]+):\S+/g, '$1');
```
gcampax, line 58: We have a function to do this used by the evaluation code...

gcampax: And also, the regular expression is wrong because parameter names can have numbers.

tool/compute-synthetic-coverage.js
```diff
@@ -0,0 +1,108 @@
...
+            help: `Path to the file containing the evaluation data (in TSV format: id, utterance, thingtalk)`
+        });
+    },
+
+    async execute(args) {
+        let syntheticSize = 0;
+        const syntheticPrograms = {}; // counter of unique programs in synthetic set
+        args.synthetic_set.setEncoding('utf8');
+        const synthetic = args.synthetic_set.pipe(csvparse({ relax: true, delimiter: '\t' }));
+        synthetic.on('data', (line) => {
```
gcampax, line 56: Btw since recently this can be written more readably as:
```
for await (const line of synthetic) {
   ...
}
```
and you don't need the waitEnd

tool/compute-training-coverage.js
```diff
@@ -0,0 +1,140 @@
...
+
+        let trainingSize = 0;
+        const trainingPrograms = new Map(); // counter of unique programs in training set
+        args.training_set.setEncoding('utf8');
+        const training = await readAllLines([args.training_set])
+            .pipe(new DatasetParser({ preserveId: true }));
+        training.on('data', async (line) => {
+            const normalized = await normalize(line.preprocessed, line.target_code, schemas);
+            const requoted = Array.from(requoteProgram(normalized)).join(' ');
+            trainingPrograms[requoted] = (trainingPrograms[requoted] || 0) + 1;
```
gcampax, line 90: This is not how you use a Map ;)

tool/compute-training-coverage.js
```diff
@@ -0,0 +1,140 @@
...
+            trainingSize += 1;
+        });
+        await waitEnd(training);
+
+        let newCount = 0;
+        const newPrograms = new Set();
+
+        let evaluationSize = 0;
+        const evaluationPrograms = {}; // counter of unique programs in evaluation set
```
gcampax, line 99: If you changed the other one to Map...

rayslxu: lol

tool/compute-training-coverage.js
rayslxu: All fixed in the last commit. (I will just do a squash merge)

PR #248, opened 2020-07-27
Prototype for evaluate-by-device

satojk:
This is the beginning of the first version of per-device-evaluation.

The device selection is currently bad, as it gives a lot of gibberish attached. One quick, hacky and imperfect solution would be to split `uniqueDevices` by commas, and select the first non "execute" bit (which could be done right after line 344 in `evaluation/sentence_evaluator.js`). This would give the right device in all cases in which there is only one device. Where there is more than one domain, it would give you the "first" device.

As per discussion in Slack, we actually want to count this sentence for each of the devices involved.

The right way to solve this is to perform this correct device extraction around line 190 instead. This would also require sending in the same "sentence" to be evaluated twice - once with each domain.

### DIFF
--- a/lib/dataset-tools/evaluation/sentence_evaluator.js
+++ b/lib/dataset-tools/evaluation/sentence_evaluator.js
@@ -108,6 +108,7 @@ class SentenceEvaluator {
             id: this._id,
             preprocessed: this._preprocessed,
             target_code: this._targetPrograms,
+            target_devices: [],
             ok: [],
             ok_without_param: [],
             ok_function: [],
@@ -186,6 +187,7 @@ class SentenceEvaluator {
         const goldFunctions = untypedTargetCode.map((code) => Array.from(getFunctions(code)));
         const goldDevices = untypedTargetCode.map((code) => Array.from(getDevices(code)));
         result.is_primitive = goldFunctions[0].length === 1;
+        result.target_devices = goldDevices;
 
         let first = true;
         let ok = false, ok_without_param = false, ok_function = false,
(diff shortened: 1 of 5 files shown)


lib/dataset-tools/evaluation/sentence_evaluator.js
```diff
@@ -330,128 +332,141 @@ class CollectSentenceStatistics extends Stream.Writable {
...
+        // _buffer will map devices to individual buffers. If splitByDevice is
+        // false, there will only be a single buffer. Othersie, these individual
+        // buffers will be created ad-hoc, as new devices come up.
+        this._buffer = {};
     }
 
     _write(ex, encoding, callback) {
-        this._buffer.total ++;
+        const uniqueDevices = ex.target_devices.filter((val, ix) => ex.target_devices.indexOf(val) === ix);
+        assert(uniqueDevices.length === 1);
```
gcampax, line 345: I don't understand what this code is trying to do?
Why do you not loop over all devices? It seems it would solve all the problems you were mentioning in the cover letter.

lib/dataset-tools/evaluation/sentence_evaluator.js
```diff
@@ -186,6 +187,7 @@ class SentenceEvaluator {
         const goldFunctions = untypedTargetCode.map((code) => Array.from(getFunctions(code)));
         const goldDevices = untypedTargetCode.map((code) => Array.from(getDevices(code)));
         result.is_primitive = goldFunctions[0].length === 1;
+        result.target_devices = goldDevices;
```
gcampax, line 190: goldDevices is only the devices in the target, not in the context, so "cancel" and similar will be accounted to no device at all. Is that the expected behavior?

lib/dataset-tools/evaluation/sentence_evaluator.js
satojk: I think you are right. I think I was too heavily anchored on the idea that this was hacky (since when I was experimenting with it, I was only getting a single string here).

The idea behind this code was that I wanted to filter out duplicates. I think this was misguided because it seems like this wasn't an array of devices, like I thought it was, but rather a singleton array with a string that had all "devices" (loosely defined - "transaction.execute()" was showing up here as a device for example. At least that's what seemed to be showing up in my tests. But you're right, a clean loop here would solve it. I'll get to it.

gcampax: It comes from goldDevices, which should be an array of device names. It will contain duplicates, but most of the examples only have one device. You can dedupe with `Array.from(new Set(...))`
Now, the problem might be in requoting.js where getFunctions and getDevices are defined. They consider all '@' tokens to be function calls, which was the case before we introduced dialogue states but is not correct now. We should fix that.

lib/dataset-tools/evaluation/sentence_evaluator.js
satojk: In my head I was including this in the bucket of "utterances without a domain". If I recall correctly you had mentioned these didn't matter much because the immediate goal was to calculate slot accuracy per domain, so I was thinking of lumping these in a "device-agnostic" category. Otherwise, I can keep a "memory" in `CollectSentenceStatistics` that remembers what the last device was, and then assign this device to these utterances. Does this seem better?

lib/dataset-tools/evaluation/sentence_evaluator.js
satojk: Looking into that! Thanks!

lib/dataset-tools/evaluation/sentence_evaluator.js
gcampax: Yeah probably better to keep them as "no domain" for now, and we can revisit later.

satojk:
I think it's ready now @gcampax .

By the way, the reason I was "getting a single string [with all domains]" in the device list was that the device list was actually a singleton list, containing.... the list of devices. so now I'm getting the first (only) element in line 190.

gcampax:
Oh yeah it's an array with one element for each valid target thingtalk, in case there are multiple acceptable interpretations. I forgot about that...

lib/dataset-tools/evaluation/sentence_evaluator.js
```diff
@@ -186,6 +187,7 @@ class SentenceEvaluator {
         const goldFunctions = untypedTargetCode.map((code) => Array.from(getFunctions(code)));
         const goldDevices = untypedTargetCode.map((code) => Array.from(getDevices(code)));
```
gcampax, line 188: Actually, are you going to fix this function so it doesn't return the dialogue policy name as a "device"?

satojk: I was hesitant to do that because I wasn't sure if this function was being used elsewhere, possibly relying on this behavior. But I can do that. I'll also look into the tests - another PR or should it go together with this one?

Failed the linter. Fixing it now.

satojk: Oh, I grep'd it and seems like it'll be fine to change this. Doing that now too

lib/dataset-tools/requoting.js
```diff
@@ -193,7 +193,9 @@ function* getFunctions(program) {
 }
 
 function* getDevices(program) {
-    for (let fn of getFunctions(program)) {
+    let functions = getFunctions(program);
+    functions.next(); // discard the policy name (the first "function")
+    for (let fn of functions) {
```
gcampax, line 198: Actually this won't work - when we do single shot, we skip the $dialogue statement. You should check that the first token is "$dialogue" and skip the next token. And while you're there, you should fix getFunctions too...

Bonus points if you add unit tests for these.

lib/dataset-tools/splitter.js
```diff
@@ -53,7 +53,7 @@ const SPLIT_STRATEGIES = {
     program: (id, sentence, program) => join(requoteProgram(program)),
     combination: (id, sentence, program) => {
         const functions = Array.from(getFunctions(program));
-        if (functions.length <= 1)
+        if (functions.length === 0)
```
satojk, line 56: getFunctions was being used here, and with the new getFunctions behavior, I think this change was necessary. But I would appreciate an extra pair of eyes, as I'm not familiar with the  rationale in this decision for train - eval split.

gcampax: No, this is not correct. getFunctions() returns the list of functions, and this function checks for one or two functions.
That fact that getFunctions() also returned the dialogue act was _the_ bug: it was never meant to do that. And indeed, it did not do that for single-shot dataset, because single-shot datasets don't have a dialogue act. So this code was correct as it was.

satojk: That makes sense. Reverting

PR #250, opened 2020-07-28
Changes to make the Thingpedia model work

gcampax:
Incomplete, but there are some evaluation fixes that will be helpful in testing, so let's merge this one first.


rayslxu:
All looks good. But is there any particular reason behind some tweaks? e.g., changes on constant number and weight. just want to know if there is some insight learned that is useful for me as well.

gcampax:
The changes for weight were trying to get more "play a song called foo" and less "play a song". Not sure it makes a huge difference.

The changes in number of constants is just about generation speed.

PR #254, opened 2020-07-30
Making the Spotify skill work

gcampax:
This is the Genie side of the work that implements [the design we discussed for Spotify](https://wiki.almond.stanford.edu/SkillDesign/Spotify).
There is quite a bit of work to get the agent to reply reasonably as well, but in the end, the basic Spotify scenarios work.

As the templates are affected (with some fixes too), a review from Ryan K is appreciated.


gcampax:
Actually, this PR is missing entity linking for secondary entities used at the first turn (e.g. artist, album, etc.), before Spotify actually works. I will need to implement it, but it will be another PR.

PR #257, opened 2020-08-01
Remove sound support from the test assistant

gcampax:
"genie assistant" is meant to be a quick way to test a new model
interactively. Supporting speech is nice, but it has heavy,
platform-specific dependencies (including the wake word stuff).

It was useful to have the code (even though it was not actually
working in practice) as guidance on how to enable almond-server,
which has real sound. Now that almond-server is up to date,
let's remove this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,8 +70,5 @@
       "thingtalk",
       "thingpedia"
     ]
-  },
-  "optionalDependencies": {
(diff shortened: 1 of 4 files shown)


gcampax:
By the way, the alternative to this PR is to actually initialize speech in "genie assistant", add snowboy as optional dep, and perhaps ship a wakeword.

PR #259, opened 2020-08-03
implemented hooks

ryachen01:


lib/engine/apps/exec_wrapper.js
```diff
@@ -129,7 +133,31 @@ class ExecWrapper extends ExecEnvironment {
...
     }
 
+    enterProcedure(procid, procname) {
+        // save the calling frame ID on the stack
+        this._procedureStack.push(this._procedureFrame);
+
+        // make a fresh ID for the new call
+        this._procedureFrameCounter++;
+        this._procedureFrame = this._procedureFrameCounter;
+    }
```
gcampax, line 143: This is the same implementation as the superclass right? We don't need to duplicate it.

lib/engine/apps/exec_wrapper.js
```diff
@@ -129,7 +133,31 @@ class ExecWrapper extends ExecEnvironment {
...
+
+    addExitProcedureHook(hook) {
+        this._hooks.push(hook);
+    }
+    async exitProcedure() {
+        // check that enter & exit are correctly paired
+        assert(this._procedureStack.length > 0);
+
+        // restore the frame ID of the caller from the stack
+        this._procedureFrame = this._procedureStack.pop();
```
gcampax, line 153: Same here, we can call `await super.exitProcedure()` and avoid code duplication.

lib/engine/apps/exec_wrapper.js
```diff
@@ -129,7 +133,31 @@ class ExecWrapper extends ExecEnvironment {
...
+        for (const hook of this._hooks)
+            await hook();
```
gcampax, line 155: We should probably empty the list of hooks after calling them, in case there is another statement.

lib/engine/apps/exec_wrapper.js
```diff
@@ -56,6 +56,10 @@ class ExecWrapper extends ExecEnvironment {
         this._trigger = null;
 
         this._execCache = [];
+        this._procedureFrameCounter = 0;
+        this._procedureFrame = 0;
+        this._procedureStack = [];
```
gcampax, line 61: Same here, right? The superclass owns these properties.

PR #261, opened 2020-08-04
Wip/cikm camera ready

rayslxu:


gcampax:
Is `/` a good character? I feel like it can legitimately appear in a sentence to separate two words...

rayslxu:
> Is `/` a good character? I feel like it can legitimately appear in a sentence to separate two words...

Backlash?

gcampax:
backlash is the escape character, it will be akwkard to write everywhere.
I'm thinking pipe instead.

PR #263, opened 2020-08-06
Add a tool to measure the entropy of a training set

gcampax:
As well as other useful statistics.

This was previously a [Python script](https://github.com/stanford-oval/thingpedia-common-devices/blob/master/scripts/measure.py) (and equivalent in genie-workdirs) but JS is significantly faster and uses less memory.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -25,6 +25,7 @@ rules:
   no-unused-vars:
     - error
     -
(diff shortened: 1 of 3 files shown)


rayslxu (approved):

gcampax:
I think this was merged already through some other PR.

PR #264, opened 2020-08-07
Make query to match entity type

ryachen01:


lib/dialogue-agent/entity-linking/entity_lookup.js
```diff
@@ -34,15 +61,41 @@ async function lookupEntity(dlg, entityType, entityDisplay, idEntities) {
...
+            const invocationTable = new Ast.Table.Invocation(null, invocation, schema);
+            const filteredTable = new Ast.Table.Filter(null, invocationTable, filter, schema);
+            const program = new Ast.Program(null, [], [],
+                [new Ast.Statement.Command(null, filteredTable, [Ast.Action.notifyAction()])], null);
+            const compiler = new AppCompiler(dlg.manager.schemas);
+            const compiled = await compiler.compileProgram(program);
+            const app = await dlg.engine.createApp(program);
+            const mainOutput = new QueueOutputDelegate();
+            const command = new RuleExecutor(dlg.engine, app, compiled.command, mainOutput);
+            command.start();
```
gcampax, line 81: Wait! You're executing this twice... If you call createApp, that also executes the app, and then you can just iterate `app.mainOutput` to get the results. See for example: https://github.com/stanford-oval/genie-toolkit/blob/master/lib/dialogue-agent/execution/statement_executor.js#L167

That way you don't have to duplicate so much code, and you don't have to break the abstraction to poke at engine internals like AppCompiler and RuleExecutor.
In fact, I would separate the code that constructs the query program (it can be in this file or it can be somewhere else) from the code executing the query program. And it would be nice to be able to go through the same paths as all other execution of ThingTalk code, because you need to select the device ID and stuff.

lib/dialogue-agent/entity-linking/entity-finder.js
```diff
@@ -64,6 +65,7 @@ function getBestEntityMatch(searchTerm, entityType, candidates) {
         score -= 0.1 * editDistance(searchTerm, candDisplay);
 
         let candTokens = candDisplay.split(' ');
+        candTokens = new Set(candTokens);
```
ryachen01, line 68: I made this change because previously if I tried to look for the album "the wall", the getBestEntityMatch function would return "The Writing's On The Wall" as the proper match instead of "The Wall" because the keyword "the" was matched twice on  "The Writing's On The Wall"

gcampax:
Is this ready for review now?

ryachen01:
Not yet. I realized there's a problem when there's filtering for multiple entity types. For example, if I search for an album by an artist, this code will match the album with an entity without considering the artist. I'm making changes so that before constructing the query we check the other filters and if an entity(artist) is a filter parameter of another entity(album), we first find the artist and pass it as a parameter for the album query

gcampax:
No, that's not the correct way to fix it.
The correct way is to express the original query as a join (or subquery), so the whole "$album by $artist" becomes the usual `@com.spotify.album(), id =~ ... && contains(artists, ...)`.
You should not "fish" for other parameters when doing entity linking.

At the same time, we don't have templates or syntax for subqueries, so we cannot fix that. So let's not touch that and we'll fix that correctly at a later time.

ryachen01:
ok then this should be ready for review

lib/dialogue-agent/dialogue-loop.js
```diff
@@ -174,24 +130,24 @@ module.exports = class DialogueLoop {
...
-            // nothing to do
-            break;
+            case 'wakeup':
+                // nothing to do
+                break;
 
-        default:
-            await this.fail();
+            default:
+                await this.fail();
```
gcampax, line 150: Please don't make unrelated whitespace changes.

lib/dialogue-agent/entity-linking/entity-finder.js
gcampax: "The Writing's On The Wall" has a pretty high edit distance compared to "The Wall" because of the number of deleted characters. I'm surprised by this change. Maybe we just need to lower the term for "the"?

gcampax: Anyway, can you add tests in test/unit/test_entities for this?

ryachen01:
The reason the wall matches with "The Writing's On The Wall"  is because for every character in edit distance the score decreases -0.1. However, whenever a word matches whether it be "the" or any other word the score gets +10 points, so with the current algorithms even with a significant edit distance if certain words match the entity will be given a pretty high score.

gcampax:
Yeah we should change the score term for "the" in that case.

test/unit/index.js
```diff
@@ -37,7 +36,6 @@ async function do_test(array) {
     }
 }
 
-
```
gcampax, line 40: Really, please, no whitespace changes on any lines you're not modifying. It breaks git-blame for no reason.

lib/dialogue-agent/entity-linking/entity_lookup.js
```diff
@@ -50,10 +52,12 @@ async function lookupEntity(dlg, entityType, entityDisplay, idEntities) {
             const dialgoueResults = dlg.dialogueState.history[0].results.results;
             let entityCandidates = [];
             for (const item of dialgoueResults) {
+                let canonical = item.value.id.display;
+                canonical = canonical.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
```
gcampax, line 56: Is this necessary? We'll do edit distance later right, so small changes should be ok.

I'm just wary of "surprise code" that makes you wonder why is this here...

ryachen01: This line is intended to remove accents. I added this because if I try to match beyonce I will end up with Mc Beyonce instead of beyoncé. Even though the edit distance is only once character beyonce doesn't match with beyoncé, so I believe the score for beyoncé would be -0.1. But with Mc Beyonce, theres a match with beyonce so the final score would be around 9.7. One alternate solution could be to increase the penalty for edit distance or decrease the reward for exact matches, but I'm not sure if that would cause any other problems. What do you think?

gcampax: We should make the "exact match" be less exact in the entity scoring computation. In general, heuristics like these fit well in the entity scorer, because entity scoring is inherently heuristic, and we have tests to ensure whatever changes we make to the entity scoring retain high precision and high recall. Changes outside the entity scoring are more dangerous, because the next person reading this code will not remember why it's happening, and why are accents unconditionally removed.

PR #270, opened 2020-08-12
More core changes to make the Thingpedia models work

gcampax:
I said in https://github.com/stanford-oval/thingpedia-common-devices/pull/156 there would not be any more changes.

I lied

This PR includes a bunch of other PRs too. I'll clean it up when it's ready for review.


languages/thingtalk/en/constants.genie
```diff
@@ -280,6 +280,8 @@ constant_Measure_dBm = {
 constant_Location = {
     'here'                    => new Ast.Value.Location(new Ast.Location.Relative('current_location'));
     !turking 'where i am now' => new Ast.Value.Location(new Ast.Location.Relative('current_location'));
+    'the area'                => new Ast.Value.Location(new Ast.Location.Relative('current_location'));
+    'the city'                => new Ast.Value.Location(new Ast.Location.Relative('current_location'));
     'home'                    => new Ast.Value.Location(new Ast.Location.Relative('home'));
```
rayslxu, line 289: "my neighborhood" for home (in eval set)

lib/i18n/american-english.js
```diff
@@ -137,39 +137,35 @@ class EnglishLanguagePack extends DefaultLanguagePack {
         if (program.isProgram && program.principal !== null)
             sentence = replaceMeMy(sentence);
 
-        if (!sentence.endsWith(' ?') && !sentence.endsWith(' !') && !sentence.endsWith(' .') && coin(0.5, rng))
-            sentence = sentence.trim() + ' .';
+        if (!sentence.endsWith(' ?') && !sentence.endsWith(' !') && !sentence.endsWith(' .')) {
+            if ((forTarget === 'user' && coin(0.5, rng)) || forTarget === 'agent')
+                sentence = sentence.trim() + ' .';
+        }
         if (forTarget === 'user' && sentence.endsWith(' ?') && coin(0.5, rng))
```
rayslxu, line 144: What does the second part of this commit mean and where is the change about it? (40eeacd, apply contractions consistently: always for "to be")

lib/languages/thingtalk/constants.js
```diff
@@ -39,7 +39,7 @@ function extractConstants(ast) {
 
     ast.visit(new class extends Ast.NodeVisitor {
         visitStringValue(value) {
-            addConstant('QUOTED_STRING', value.value, value);
+            addConstant('QUOTED_STRING', value.value.replace(/[ \t\v\r\n]+/g, ' ').trim(), value);
```
rayslxu, line 42: why is this a problem? don't we want to at least preserve the linebreaks?

languages/thingtalk/en/constants.genie
```diff
@@ -39,6 +39,17 @@ constant_Number = {
...
+    'three'  => new Ast.Value.Number(3);
+    'four'  => new Ast.Value.Number(4);
+    'five'  => new Ast.Value.Number(5);
+    'six'  => new Ast.Value.Number(6);
+    'seven'  => new Ast.Value.Number(7);
+    'eight'  => new Ast.Value.Number(8);
+    'nine'  => new Ast.Value.Number(9);
+    'ten'  => new Ast.Value.Number(10);
+    'eleven'  => new Ast.Value.Number(11);
+    'twelve'  => new Ast.Value.Number(12);
```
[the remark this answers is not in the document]
gcampax: Actually this is wrong: small numbers between 2 and 12 are normalized to digits so we don't see them in words. I revert this commit later on.

lib/i18n/american-english.js
gcampax: Contractions are "you are" vs "you're". Previously, we'd generate one or the other randomy. Now:
- for the user, we still do randomly (because we want to train robustly)
- for the agent, we always contract "to be" -> "i'm", "you're", "he's", "it's", "she's"
- for the agent, we never contract "would" or "have" -> "i would", "she has", "we have"

lib/languages/thingtalk/constants.js
gcampax: This is used to make the output sentence from the agent. The string will be in the middle of some larger sentence. Visually,  a line break looks out of place, and of course, there is no line break in speech...

PR #274, opened 2020-08-12
entity-retriever.js: Small bug fix

Mehrad0711:


lib/i18n/tokenizer/base.js
```diff
@@ -470,4 +470,9 @@ module.exports = class BaseTokenizer {
 
         return { tokens, rawTokens, entities };
     }
+
+    detokenize(tokens) {
+        return tokens.join(' ');
+    }
```
gcampax, line 476: I'm surprised this works, because there is already a "detokenize" function in this class.

lib/languages/thingtalk/entity-retriever.js
```diff
@@ -81,7 +81,7 @@ module.exports = class GenieEntityRetriever extends NNSyntax.EntityRetriever {
             this._sentenceContainsNonConsecutive(entityTokens) :
             this._sentenceContains(entityTokens);
         if (found)
-            return entityTokens.join(' ');
+            return this._tokenizer.detokenize(entityTokens);
```
gcampax, line 84: No, the program should contain the tokenized version of the entity, otherwise you cannot match it and requote it. You should be the first to know that!

Mehrad0711: wait this function doesn't really detokenize. It just puts tokenized words back together with the right word boundary to form the final string.
Problem at hand is with Chinese where "沃克沃斯" is tokenized to "沃 克 沃 斯" and returned that way, so params no longer match.

gcampax: I don't understand. Both the sentence and the program are tokenized identically, so they match by definition.

Mehrad0711: If you pass --no-tokenized then yes. If not, the program still gets tokenized while the sentence won't.

gcampax: Huh? If you pass --tokenized (or don't pass anything) the sentence is already tokenized, params and everything. You have a bug somewhere else.

Mehrad0711: Let's take this offline :))

gcampax:
Closing this one

PR #276, opened 2020-08-13
bug fix in entity matching

ryachen01:
Single character tokens were causing some issues with entity matching.  Maroon 5 was matching with Karaoke - Maroon 5 because 5 matches with - and 5.

### DIFF
--- a/lib/dialogue-agent/entity-linking/entity-finder.js
+++ b/lib/dialogue-agent/entity-linking/entity-finder.js
@@ -70,7 +70,7 @@ function getBestEntityMatch(searchTerm, entityType, candidates) {
         for (let candToken of candTokens) {
             let found = false;
             for (let token of searchTermTokens) {
-                if (editDistance(token, candToken) <= 1 && token.length > 1 || editDistance(token, candToken) === 0) {
+                if (token === candToken || (editDistance(token, candToken) <= 1 && token.length > 1)) {
                     score += 10;
                     found = true;
                 } else if (candToken.startsWith(token)) {
--- a/lib/dialogue-agent/entity-linking/entity-finder.js
+++ b/lib/dialogue-agent/entity-linking/entity-finder.js
@@ -56,10 +56,11 @@ function getBestEntityMatch(searchTerm, entityType, candidates) {
     let best = undefined,
         bestScore = undefined;
 
+    searchTerm = _removeParenthesis(searchTerm);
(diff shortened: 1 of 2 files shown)


lib/dialogue-agent/entity-linking/entity-finder.js
```diff
@@ -70,7 +70,7 @@ function getBestEntityMatch(searchTerm, entityType, candidates) {
         for (let candToken of candTokens) {
             let found = false;
             for (let token of searchTermTokens) {
-                if (editDistance(token, candToken) <= 1) {
+                if (editDistance(token, candToken) <= 1 && token.length > 1 || editDistance(token, candToken) === 0) {
```
gcampax, line 73: editDistance() == 0 can be written a lot faster as `token === candToken`

gcampax: Also, for clarity please put parenthesis around the first clause of ||.
Also, for speed you should always write your conditions from faster to slower, because logic operators are short-circuited.

lib/dialogue-agent/entity-linking/entity-finder.js
```diff
@@ -96,4 +97,8 @@ function getBestEntityMatch(searchTerm, entityType, candidates) {
     return best;
 }
 
+function _removeParenthesis(str){
+    return str.replace(/ \([\s\S]*?\)/g, '');
```
gcampax, line 101: \s is whitespace, and \S is everything but whitespace, so this will match literally everything? Should write `/ \(.*?\)/` instead? To avoid confusing someone who doesn't know RegExp syntax by heart (I had to check MDN...)

lib/dialogue-agent/entity-linking/entity-finder.js
```diff
@@ -96,4 +97,8 @@ function getBestEntityMatch(searchTerm, entityType, candidates) {
     return best;
 }
 
+function _removeParenthesis(str){
```
gcampax, line 100: By convention, only private methods and private fields of classes start with _. Functions should not start with _

lib/dialogue-agent/entity-linking/entity-finder.js
gcampax: Wait did you revert the fix for this in some later comment? Please check that you didn't make mistakes in rebasing this PR. Maybe do a full rebase and squash, force push and look at it before merging.

PR #277, opened 2020-08-13
starter/schemaorg: pass a lower batch size to auto-annotate during tests

gcampax:
This reduces the chances of an out-of-memory in Travis.


gcampax:
This doesn't actually have the effect we want, we still die of OOM.

PR #281, opened 2020-08-21
minor tweaks to improve agent response

ryachen01:
You can see the changes to the agent response at https://github.com/stanford-oval/thingpedia-common-devices/pull/159/commits/fd5ed3b7e77828e05c4c1280bd291a3034ce3a41

### DIFF
--- a/languages/thingtalk/slot_bag.js
+++ b/languages/thingtalk/slot_bag.js
@@ -103,16 +103,9 @@ function checkAndAddSlot(bag, filter) {
             return null;
         const clone = bag.clone();
         if (clone.has(filter.name))
-            clone.get(filter.name).value.push(filter.value);
+            return null;
         else
             clone.set(filter.name, new Ast.Value.Array([filter.value]));
-        var values = new Set();
-        for (const value of (clone.get(filter.name).value)){
-            if (values.has(value.value))
-                return null;
-            else
-                values.add(value.value);
-        }
         return clone;
     } else {
         if (filter.operator !== '==' && filter.operator !== '=~')
--- a/languages/thingtalk/slot_bag.js
+++ b/languages/thingtalk/slot_bag.js
@@ -106,6 +106,13 @@ function checkAndAddSlot(bag, filter) {
             clone.get(filter.name).value.push(filter.value);
         else
             clone.set(filter.name, new Ast.Value.Array([filter.value]));
+        var values = new Set();
+        for (const value of (clone.get(filter.name).value)){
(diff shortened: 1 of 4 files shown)


languages/thingtalk/slot_bag.js
```diff
@@ -106,6 +106,13 @@ function checkAndAddSlot(bag, filter) {
             clone.get(filter.name).value.push(filter.value);
         else
             clone.set(filter.name, new Ast.Value.Array([filter.value]));
+        var values = new Set();
+        for (const value of (clone.get(filter.name).value)){
+            if (values.has(value.value))
+                return null;
+            else
+                values.add(value.value);
+        }
```
gcampax, line 115: This code doesn't work for any value that is not a JS primitive (number, boolean, string, enum). This includes TT entities, dates, times, measurements.
What is this code trying to do exactly? (A comment wouldn't hurt here)

ryachen01: I added this to try to get rid of the by {artist} by {artist} by {artist} problem. It's supposed to check the filter parameters and make sure that none of them are the same. Would this approach work if extended to the other types or would a different approach be better?

gcampax: So, one thing I don't quite understand is why checking the name of the slot is not sufficient. A SlotBag is basically a map from name to value. If you have two slots called "artist" the function returns `null` already. So what is going on here?

ryachen01: I check value because some songs have multiple artists, so since right now there aren’t templates for song by artist 1 and artist 2 by only returning null when the values are the same that still allows the agent to return something like song by artist 1 and by artist 2. I guess it would probably be better to not change this function until the appropriate template is added.

gcampax: Ugh I forgot about multiple artists. That's complicated. Still, that does not explain why there can be multiple "by X" for songs with only one artist.

ryachen01: Yeah, I'm not 100% sure why that happens either, but replacing line 106 with return null solves the problem of multiple by X by ensuring that only one artist slot can be added

gcampax: Oh I see what's going on here! This code is too smart for its own good, and tries to build the array on the fly. That's not correct, as you note, because one needs a proper template for arrays instead of just appending the same slot twice. So let's replace line 106 with a null and be done with this.

PR #282, opened 2020-08-23
All changes made for cikm camera ready

rayslxu:


tool/autoqa/schemaorg/manual-annotations.js
```diff
@@ -479,12 +486,21 @@ const ENUM_VALUE_NORMALIZE = {
     }
 };
 
+const WHITELISTED_PROPERTIES_BY_DOMAIN = {
```
gcampax, line 607: It's low on the list of things to worry about, but we should try to move away from blacklist/whitelist as terminology. It is not a great choice of words, on many levels.

tool/autoqa/schemaorg/trim-class.js
```diff
@@ -172,6 +175,8 @@ class SchemaTrimmer {
 
         for (let fieldname in type.fields) {
             const field = type.fields[fieldname];
+            if (this._whiteListed(`${arg.name}.${fieldname}`))
```
gcampax, line 189: Is there a case here where you can have two level nesting of compound types?

languages/thingtalk/ast_manip.js
```diff
@@ -615,9 +615,12 @@ function checkAtomFilter(table, filter) {
     ptype = table.schema.out[filter.name];
     vtype = ptype;
     if (filter.operator === 'contains') {
-        if (!vtype.isArray)
+        if (ptype.isArray)
+            vtype = vtype.elem;
+        else if (ptype.isRecurrentTimeSpecification)
+            vtype = Type.Date;
```
gcampax, line 644: Date or Time right?

languages/thingtalk/common-constants.genie
```diff
@@ -45,7 +45,7 @@ constant_Currency = const(CURRENCY, Type.Currency);
 constant_Time = const(TIME, Type.Time);
 constant_Date = const(DATE, Type.Date);
 constant_Location = const(LOCATION, Type.Location);
-
+constant_RecurrentTimeSpecification = const(RECURRENT_TIME_SPECIFICATION, Type.RecurrentTimeSpecification);
```
gcampax, line 48: We don't have a RECURRENT_TIME_SPECIFICATION token, so this is not correct?

languages/thingtalk/utils.js
```diff
@@ -53,7 +53,10 @@ function makeFilter(loader, pname, op, value, negate = false) {
     if (ptype.isEntity && ptype.type === 'tt:url')
         return null;
     if (op === 'contains') {
-        ptype = Type.Array(vtype);
+        if (loader.params.out.has(pname.name + '+' + Type.RecurrentTimeSpecification))
+            ptype = Type.RecurrentTimeSpecification;
+        else
+            ptype = Type.Array(vtype);
```
gcampax, line 59: This is a little bit bad, once you have multiple domains using different types for the same parameter name.

tool/autoqa/schemaorg/manual-annotations.js
```diff
@@ -189,11 +189,14 @@ const MANUAL_PROPERTY_CANONICAL_OVERRIDE = {
     // restaurants
     'datePublished': {
         passive_verb: ["published | on #", "written | on #"],
-        base: ["date published"]
+        base: ["date published"],
+        adjective_argmax: ['most recent', 'latest', 'last'],
+        adjective_argmin: ['earliest']
```
gcampax: Lol we did not describe these in the paper...
Also, "first", "oldest".

tool/lib/constants-sampler.js
```diff
@@ -114,6 +114,12 @@ module.exports = class ConstantSampler {
                                 constants.push([`param:@${device}.${f}:${arg}:Entity(${entityType})`, `null`, sample]);
                             });
                         }
+
+                        if (arg === 'id') {
+                            samples.forEach((sample) => {
+                                constants.push([`param:@${device}.${f}:${arg}:String`, sample]);
+                            });
+                        }
```
gcampax, line 122: What is the bug exactly here? (That is, can you expand the commit message?)

languages/thingtalk/en/constants.genie
```diff
@@ -44,21 +44,32 @@ constant_date_point = {
...
-            'next month'          => C.makeDate(new Ast.DateEdge('end_of', 'mon'), '+', null);
-            'next year'           => C.makeDate(new Ast.DateEdge('end_of', 'year'), '+', null);
+            'the end of the month'=> C.makeDate(new Ast.DateEdge('end_of', 'mon'), '+', null);
+            'the end of the year' => C.makeDate(new Ast.DateEdge('end_of', 'year'), '+', null);
+        }
+    }
+}
+
+constant_date_range = {
+    'this week'   => [C.makeDate(new Ast.DateEdge('start_of', 'week'), '+', null), C.makeDate(new Ast.DateEdge('end_of', 'week'), '+', null)];
```
gcampax, line 86: Does this mean we annotate "news published this week" with two comparisons?

languages/thingtalk/utils.js
```diff
@@ -86,6 +86,16 @@ function makeAndFilter(loader, param, op, values, negate=false) {
     return f;
 }
 
+function makeDateRangeFilter(loader, param, values) {
+    if (values.length !== 2)
+        return null;
+    const operands = [
+        makeFilter(loader, param, '>=', values[0]),
+        makeFilter(loader, param, '<=', values[1])
+    ];
```
gcampax, line 95: Check if either filter is null ?

languages/thingtalk/load-thingpedia.js
```diff
@@ -834,6 +837,86 @@ class ThingpediaLoader {
...
+        expansion = [new this._runtime.NonTerminal('constant_Number'), 's']; // 90s
+        this._grammar.addRule('constant_date_range', expansion, this._runtime.simpleCombine((value, _2) => [
+            makeDate(new Ast.DatePiece(value.value, null, null, null), '+', null),
+            makeDate(new Ast.DatePiece(value.value, null, null, null), '+', new Ast.Value.Measure(10, 'year'))
+        ]));
+
+        // year + month
+        for (let month = 1; month <= 12; month ++) {
+            for (let phrase of months[month-1]) {
+                expansion = [phrase, 'of', new this._runtime.NonTerminal('constant_Number')];
```
gcampax: If we care about translation (and I think we do), we really cannot do this loop stuff. Why do you even need it anyway? Why are month and day not just a plain non-terminal?

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -520,6 +527,9 @@ module.exports = class ParameterReplacer {
                 varref.token = token;
                 entities[slot] = varref;
                 out.push(slot);
+            } else if (/^[A-Z]+_[0-9]+/.test(token)) {
+                entities[token] = makeDummyEntity(token);
+                out.push(token);
```
gcampax, line 545: This change is in the wrong commit (if you care about that)

languages/thingtalk/en/parameters.genie
```diff
@@ -116,6 +116,36 @@ where_question = {
...
+    'is there' ('any' | '') table:with_filtered_table ('that is'|'') p:npi_boolean_projection =>  C.makeSingleFieldProjection('table', Type.Boolean, table, p.name);
+
+    // npv
+    'is' ('the' | '') table:with_filtered_table p:npv_boolean_projection => C.makeSingleFieldProjection('table', Type.Boolean, table, p.name);
+    'is there' ('any' | '') table:with_filtered_table ('that is'|'') p:npv_boolean_projection =>  C.makeSingleFieldProjection('table', Type.Boolean, table, p.name);
+
+    // apv
+    'is' ('the' | '') table:with_filtered_table p:apv_boolean_projection => C.makeSingleFieldProjection('table', Type.Boolean, table, p.name);
+    'is there' ('any' | '') table:with_filtered_table ('that is'|'') p:apv_boolean_projection =>  C.makeSingleFieldProjection('table', Type.Boolean, table, p.name);
+}
```
gcampax, line 127: How does this compare with "boolean_user_search_question" (which is a coref question)?

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -341,6 +341,14 @@ module.exports = class ParameterReplacer {
                     sentenceValue = sentenceValue.substring(match[0].length);
             }
         }
+
+        // hack for hotel domain: remove "hotel" in the end of names
+        if (arg.name === 'id' && programValue.endsWith(' hotel')) {
+            programValue = programValue.substring(0, programValue.length - ' hotel'.length);
+            if (coin(0.5, this._rng))
+                sentenceValue = sentenceValue.substring(0, sentenceValue.length - ' hotel'.length);
+        }
```
gcampax, line 355: Should probably do this only if the domain is hotel. Otherwise, searching for [Tokyo Hotel](https://en.wikipedia.org/wiki/Tokio_Hotel) won't work very well...
(Used to be all the rage for my female friends when I was in middle school!)

lib/i18n/american-english.js
```diff
@@ -270,6 +270,16 @@ class EnglishLanguagePack extends DefaultLanguagePack {
         return [inflected, ...words.slice(1)].join(' ');
     }
 
+    // to verb base
+    toVB(verb) {
```
gcampax, line 274: please use long names for the methods, and also please use jsdoc syntax for documentation

languages/thingtalk/ast_manip.js
rayslxu: fixed.

languages/thingtalk/utils.js
rayslxu: Not sure what's the best soln, but I added another constaint checking if vtype is date or time

tool/autoqa/schemaorg/trim-class.js
rayslxu: fixed

languages/thingtalk/common-constants.genie
rayslxu: dropped.

languages/thingtalk/en/constants.genie
rayslxu: yes, e.g.,  "show me meetings this week" (but in this case we want range between now and end of week, instead of start of week and end of week.....)

gcampax: That's a tricky one. We'll need a lot of annotations.

tool/autoqa/schemaorg/manual-annotations.js
rayslxu: added

tool/lib/constants-sampler.js
rayslxu: We use string value instead of entity value when filtering on id...

languages/thingtalk/utils.js
rayslxu: added the check

lib/dataset-tools/augmentation/replace_parameters.js
rayslxu: Addressed

tool/lib/constants-sampler.js
gcampax: Oh I see.

languages/thingtalk/load-thingpedia.js
rayslxu: this is already moved to `constants.genie`, loops just makes it slightly easier to write...

gcampax: Easier to write, maybe. But not easier to read, and that's more important. It would be a lot more readable with a non-terminal for "day", a non-terminal for "month", and proper grammar to combine the two. Plus it won't work with my translation PR to have a loop.

languages/thingtalk/en/parameters.genie
rayslxu: These two are still for unknown entities. the second one is clearer as it is just doing a search. the first one is to generate things like "is the restaurants in palo alto Chinese".

gcampax: No yeah the search stuff I understand, that's not projection and the grammar is different (it would match "negative_proposal_reply" from the refinement templates). I just wonder if we can unify the truly boolean templates, at least grammar-wise. Anyway not important right now.

languages/thingtalk/load-thingpedia.js
rayslxu: https://github.com/stanford-oval/genie-toolkit/blob/wip/cikm-camera-ready/languages/thingtalk/en/constants.genie#L146
this is what i have right now. why this cannot be translated? is it because of `#(phrase)`?

gcampax: Yes. It makes the templates dynamic, so we cannot statically extract all the templates and prepare a translation file. (You should take a look at #285 which I just pushed.

rayslxu: got it. I will fix this.

rayslxu: fixed in 2b3515b

lib/i18n/american-english.js
rayslxu: fixed in d35c0d1

PR #283, opened 2020-08-23
Pass thingtalk code to nlp server

Mehrad0711:


lib/prediction/predictor.js
```diff
@@ -63,6 +63,8 @@ class Worker extends events.EventEmitter {
         ];
         if (process.env.GENIENLP_EMBEDDINGS)
             args.push('--embeddings', process.env.GENIENLP_EMBEDDINGS);
+        if (process.env.DATABASE)
```
gcampax, line 66: This env variable should be prefixed, "database" is too generic of a word.

lib/prediction/predictor.js
```diff
@@ -107,7 +109,7 @@ class Worker extends events.EventEmitter {
         this._requests.clear();
     }
 
-    request(task, context, question) {
+    request(task, context, question, answer=undefined) {
```
gcampax, line 112: You don't need `=undefined`, in JS everything is undefined if unspecified.
(And if you want it, put spaces before and after the equal sign)

lib/prediction/predictor.js
```diff
@@ -119,13 +121,13 @@ class Worker extends events.EventEmitter {
...
         assert(typeof question === 'string');
-        this._stream.write({ id, context, question, task });
+        this._stream.write({ id, context, question, answer, task });
         return promise;
     }
 }
 
 module.exports = class Predictor {
-    constructor(id, modeldir, nWorkers = 1) {
+    constructor(id, modeldir, nWorkers= 1) {
```
gcampax, line 130: No whitespace changes.

tool/evaluate-server.js
```diff
@@ -162,7 +162,12 @@ module.exports = {
             required: false,
             type: Number,
             default: 0,
-            help: 'Collapse all examples of complexity smaller or equal to this'
+            help: 'Collapse all examples of complexity smaller or equal to this',
+        });
+        parser.addArgument('--oracle', {
```
gcampax, line 167: We recently updated argparse, the new API is called `add_argument` (it will show a deprecation warning otherwise)

lib/prediction/predictor.js
```diff
@@ -63,6 +63,8 @@ class Worker extends events.EventEmitter {
         ];
         if (process.env.GENIENLP_EMBEDDINGS)
             args.push('--embeddings', process.env.GENIENLP_EMBEDDINGS);
+        if (process.env.DATABASE)
+            args.push('--entity_database', process.env.ENTITY_DATABASE);
```
gcampax, line 67: Uhm no? The if statement is wrong. Also, when I say prefixed, I mean prefixed with GENIENLP.

PR #284, opened 2020-08-25
Fix wikidata unit test

rayslxu:
Apparently wikidata is actively growing, relax the assert statement


rayslxu:
> This probably works, but I would prefer to include a snapshot of whatever the API returns and parse that instead of making a live API call in a unit test. It's a unit test, it should test only a single piece of code.

If we want to mock the output, then there is no point for this test really... I will merge to unblock you, we may just drop this test entirely in the future if you want.

gcampax:
The point of the test is to test the wikidata code, isn't it? I don't understand why there would be no point in mocking.

PR #285, opened 2020-08-27
Translation support for templates: a new approach

gcampax:
This is a Gettext-based approach to translating templates.
We run static code extraction on the templates, prepare a file for translation, translate using standard tools people are familiar with, and then merge back into the templates.

The last bit is still missing so this PR is a draft.


gcampax:
This was superseded by #542

PR #322, opened 2020-09-09
Another round of bug fixes

gcampax:
Fixes some of the issues reported in the last rounds of testing.


lib/i18n/american-english.js
```diff
@@ -300,6 +300,22 @@ class EnglishLanguagePack extends DefaultLanguagePack {
...
+            case 'my':
+                return 'your';
+            case 'mine':
+                return 'yours';
+            default:
+                return what;
+            }
+        });
+    }
+
```
rayslxu, line 318: Why is this not turning "Alright, I shuffled your Spotify." to "Alright, you shuffled your Spotify."

gcampax: Because we only apply it to the canonical forms in Thingpedia. The words in the templates are unchanged.
This is also why I need to do it this way instead of adding to the postprocessing in the sentence.

rayslxu (approved):

PR #347, opened 2020-09-12
Add support for new thingtalk units: area and volume

rayslxu:


gcampax (approved):
Yeah let's get this in

PR #349, opened 2020-09-13
dialogue-agent: reset the state after a successful action

gcampax:
That way, the interpretation of subsequent command will not be
affected by the previous ones and will be more correct.

This is the "big hammer" of reducing dialogues to the minimum.
With this, Spotify is basically free of all dialogues, unless
it's Q&A.

---

I'm not sure about this PR, but Monica seems to want it?


rayslxu (commented):
I thought we can somewhat fix this by tuning synthetic data, something like stop/restart the conversation once an action succeeded with some probability.

gcampax:
We already do that (no probability, we just always try to continue the conversation) but it doesn't seem to help. It's possible that the space of conversations is too big, or it's possible that the simulator is not faithful enough to the real thing.

rayslxu (approved):

PR #357, opened 2020-09-16
Pass over the Wikidata starter code

gcampax:


rayslxu (commented):
We might need to cut things for the homework, but overall look good! Thanks. 

one thing missing is that we need to show them how to edit MANUAL_PROPERTY_CANONICAL_OVERRIDE in manual annotations, probably with a few examples. 

also, there is a typo (paraprase)

gcampax:
Look at https://github.com/stanford-oval/cs294-fall2020/pull/1 for the homework. I refer to specific steps of the starter that they need to follow.

rayslxu (approved):
let's merge it, and deal with the other problems later.

PR #361, opened 2020-09-24
Add "new" in front of all Type constructors

gcampax:
As part of the TypeScript-ification of ThingTalk, the Type
class is no longer an adt, and is now a regular class. Hence,
we need to use "new" to construct non-singleton types.

Let's do this early so nothing will break when we merge
this change in ThingTalk.


rayslxu (approved):
looks good. You did a grep to make sure everything is included, right?

PR #374, opened 2020-10-09
Some updates from SPL work 

Mehrad0711:
Most changes are around better handling of string parameters for augmentation.
Also, updating requote and augmentation scripts to accept examples with multiple TT annotations.


tool/autoqa/schemaorg/manual-annotations.js
```diff
@@ -563,12 +563,12 @@ const STRUCT_INCLUDE_THING_PROPERTIES = new Set([
...
 
 const STRING_FILE_OVERRIDES = {
-    'org.schema.Restaurant:Restaurant_name': 'org.openstreetmap:restaurant',
+    // 'org.schema.Restaurant:Restaurant_name': 'org.openstreetmap:restaurant',
     'org.schema.Person:Person_name': 'tt:person_full_name',
     'org.schema.Person:Person_alumniOf': 'tt:university_names',
     'org.schema.Person:Person_worksFor': 'tt:company_name',
     'org.schema.Person:Person_jobTitle': 'tt:job_title',
-    'org.schema.Hotel:Hotel_name': 'org.openstreetmap:hotel',
+    // 'org.schema.Hotel:Hotel_name': 'org.openstreetmap:hotel',
```
gcampax: What's special about Restaurant and Hotel? What about all the other domains? Your scripts should remove `#[string_values]` or replace them with a localized file instead.

tool/requote.js
```diff
@@ -51,9 +51,10 @@ function findSpanContaining(index, spansBySentencePos) {
     return false;
 }
 
+// check normalized version of tokens to handle accent marks/ grapheme properly
 function substringIsFound(sequence, substring, i) {
     for (let j = 0; j < substring.length; j++) {
-        if (sequence[i+j] !== substring[j])
+        if (sequence[i+j].normalize() !== substring[j].normalize())
```
gcampax, line 57: This code is run on tokenized examples (sentence/programs) so everything is NFKD already. Why is this needed?

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -106,12 +106,13 @@ class UniformValueList {
...
 
 class ValueListLoader {
-    constructor(provider, samplingType, rng) {
+    constructor(provider, samplingType, subsetParamSet, rng) {
         this._provider = provider;
 
         this._emptyList = new UniformValueList([]);
         this._cache = new Map;
         this._samplingType = samplingType;
+        this._subsetParamSet = subsetParamSet;
```
gcampax, line 115: This option should not be a string, it should be a tuple of numbers. Do the parsing outside.

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -425,6 +425,11 @@ module.exports = class ParameterReplacer {
                 continue;
             }
 
+            if (sampled === 'unspecified'){
+                attempts -= 1;
+                continue;
+            }
```
gcampax, line 439: What's this about? No.

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -425,7 +425,7 @@ module.exports = class ParameterReplacer {
             }
 
             // fail immediately if sampled value includes return char
-            if (sampled.includes('\n')){
+            if (sampled.includes('\n') || sampled.includes('/') || sampled.includes('\\') ){
```
gcampax, line 431: What's going on here? Sampled is a tokenized value so it cannot have `\n`. And why is `/` or `\\` problematic? We certainly want to understand words with slashes.

tool/evaluate-server.js
```diff
@@ -175,6 +175,12 @@ module.exports = {
             default: 0,
             help: 'Start evaluation from this line of input data',
         });
+        parser.add_argument('--output-errors', {
+            required: false,
+            default: process.stdout,
+            type: fs.createWriteStream,
+            help: "Write erred examples to this file instead of stdout"
```
gcampax, line 182: Btw "erred" is not a word, the word is "erroneous" (or "bogus" which is more common)

gcampax (changes requested):
A lot of comments throughout but overall it's ok. I really don't like the option to skip bogus examples for evaluation though. It's way too easy to confuse "errors" with parsing errors, and there is no reason to have invalid examples in the evaluation set.

lib/dataset-tools/augmentation/replace_parameters.js
Mehrad0711: For some languages, the tokenization was wrong for parameters with slashes in them. I will remove this and revisit later if the problem came up again.

gcampax: We don't split on slashes, but we do split on backslashes. At least in English. I think other languages do the same because they don't have custom rules.

tool/augment.js
```diff
@@ -155,6 +155,12 @@ module.exports = {
             help: 'Random/ Uniform strategy assigns random/ uniform weights to parameters instead of reading from file',
             default: false
         });
+        parser.add_argument('--subset-param-set', {
+            required: false,
+            type: Array,
```
gcampax, line 160: This type notation is incorrect. `type` is a function that takes each value and returns the parse. You're passing one string value, and converting to array, you will get one element. Did you test this?

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -330,7 +330,10 @@ module.exports = class ParameterReplacer {
     _transformValue(sentenceValue, programValue, arg) {
         if (this._requotable) {
             if (arg && arg.name === 'id' && arg.type.isEntity && arg.type.type.endsWith(':Hotel')) {
-                sentenceValue = sentenceValue.slice(0, sentenceValue.indexOf(',')).trim();
+                if (sentenceValue.startsWith('the '))
+                    sentenceValue = sentenceValue.slice('the '.length).trim();
```
gcampax, line 334: You should use the "definite article" stuff from langPack instead of this.

lib/dataset-tools/augmentation/replace_parameters.js
```diff
@@ -330,7 +330,10 @@ module.exports = class ParameterReplacer {
...
+                if (sentenceValue.indexOf(',') !== -1)
+                    sentenceValue = sentenceValue.slice(0, sentenceValue.indexOf(',')).trim();
```
gcampax, line 336: This does not seem correct in general. Also, `,` is not a good word so it will never show up.

gcampax (changes requested):
A few more comments.

tool/augment.js
Mehrad0711: Ok, I changed it so we that we pass a string to augment-worker, and then parse it to an array before passing to the dataset augmenter.

lib/dataset-tools/augmentation/replace_parameters.js
Mehrad0711: The "the" bug was coming from English side caused; where requoting was capturing "casablanca" instead of "the casablanca" for example and thus you ended up with excessive "the" everywhere. The bug is fixed and this is for cleanups on datasets generated with that bug.

lib/dataset-tools/augmentation/replace_parameters.js
Mehrad0711: American English GoodWord accepts commas as well so that we can have locations  like "washington , dc" in params.
Some values from TripAdvisor contained repetitive/ excessive fields like " city, city, country" or "area, city, town, country". This is to clean those up.

tool/augment.js
gcampax: That's fine

lib/dataset-tools/augmentation/replace_parameters.js
gcampax: I don't quite understand.

lib/dataset-tools/augmentation/replace_parameters.js
gcampax: It's not a good heuristic to do for everything then. Put it in make-string-datasets

lib/dataset-tools/augmentation/replace_parameters.js
Mehrad0711: his is not for removing lang specific articles from the beginning of parameters. Same token ("the") is repeated regardless of language. I will just do this processing outside of genie and remove this here. That would create less confusion I guess.

gcampax: Yes please.

gcampax (approved):

PR #376, opened 2020-10-12
Wip/emnlp autoqa camera ready

rayslxu:


tool/autoqa/schemaorg/manual-annotations.js
```diff
@@ -183,6 +183,10 @@ const PROPERTY_CANONICAL_OVERRIDE = {
         base: ['location', 'address'],
         preposition: ["in #", "from #", "around #", "at #", "on #"]
     },
+    'postalCode': {
+        base: ['postal code', 'postcode', 'zip code'],
+        preposition: ['in #', 'from #'],
```
gcampax, line 194: `in the # zip code` perhaps?

tool/autoqa/lib/canonical-extractor.js
```diff
@@ -27,10 +27,13 @@ const child_process = require('child_process');
...
-    "2": "two",
-    "3": "three",
-    "4": "four"
+    "1": ["one"],
+    "2": ["two"],
+    "3": ["three"],
+    "4": ["four"],
+    "feb 14 2017": ["february 14 2017", "february 14, 2017", "feb 14, 2017", "14 february 2017", "14 february, 2017", "14 feb 2017", "14 feb, 2017"],
+    "may 4th, 2016": ["may 4 2016", "may 4, 2016", "may 4th 2016", "4th may 2016", "4th may, 2016", "4 may 2016", "4 may, 2016"],
+    "august 2nd 2017": ["august 2, 2017", "august 2 2017", "august 2nd, 2017", "2 august 2017", "2nd august 2017", "2 august, 2017", "2nd august, 2017"]
```
gcampax, line 37: Is it worth using the tokenizer to extract these? We can already identify dates pretty well.

tool/autoqa/schemaorg/process-schema.js
```diff
@@ -318,6 +318,8 @@ class SchemaProcessor {
                 metadata.counted_object = MANUAL_COUNTED_OBJECT_OVERRIDE[propertyname];
             else if (propertyname.startsWith('numberOf'))
                 metadata.counted_object = [ clean(propertyname.slice('numberOf'.length)) ];
+            else if (propertyname.startsWith('num') && propertyname.charAt(3) === propertyname.charAt(3).toUpperCase())
```
gcampax, line 321: Writing this as a regular expression would be a lot readable...

tool/autoqa/schemaorg/manual-annotations.js
```diff
@@ -351,6 +351,10 @@ const MANUAL_PROPERTY_CANONICAL_OVERRIDE = {
         verb_projection: ['know', 'understand', 'master'],
         adjective: ['# speaking']
     },
+    addressLocality: {
+        base: ['city'],
```
gcampax, line 355: city, town, area, etc.

tool/autoqa/lib/bert-canonical-annotator.py
```diff
@@ -212,6 +213,10 @@ def predict_one(self, table, arg, query, word, k):
         :param k: number of top candidates to return, this defaults to self.k if absent
         :return: a array in length k of predicted tokens
         """
+        # skip uncommon word
+        if not wordnet.synsets(word):
```
gcampax: Is it worth to bring in all of nltk and the wordnet dataset just to learn that a word is common or not? Seems like an overkill...

tool/autoqa/auto-annotate.js
```diff
@@ -68,19 +68,14 @@ module.exports = {
...
+            help: 'Remove all existing canonical annotations in the schema',
             default: false
         });
-        parser.add_argument('--algorithm', {
-            help: 'Different algorithms to generate canonicals including bert, bart, adj, domain_synonyms split by comma (no space)',
-            default: null
+        parser.add_argument('--algorithms', {
+            nargs: '*',
+            help: 'Different algorithms to generate canonicals including bert-property-synonyms, bart-paraphrase, bert-adjectives, bert-domain-synonyms'
         });
```
gcampax, line 79: What's the default here? Should we have a reasonable default? Is that BART only?

starter/schemaorg/Makefile
```diff
@@ -91,13 +99,13 @@ custom_train_nlu_flags ?=
...
 .SECONDARY:
 
-models/paraphraser-bart:
-	mkdir models
-	wget --no-verbose https://almond-static.stanford.edu/test-data/paraphraser-bart.tar.xz
-	tar -C models -xvf paraphraser-bart.tar.xz
+models/paraphraser-bart-large-speedup-megabatch-5m:
+	mkdir -p models
+	wget --no-verbose https://almond-static.stanford.edu/test-data/paraphraser-bart-large-speedup-megabatch-5m.tar.xz
+	tar -C models -xvf paraphraser-bart-large-speedup-megabatch-5m.tar.xz
```
gcampax, line 105: Do we need to distribute both paraphraser models? Can we kill the old one?

gcampax (commented):
Most of the changes are in autoqa's canonical generation, so I only gave a cursory look, cause I'm not very familiar with the code. If you want I can read more carefully, but I trust your tests really. I left a few comments here and there.

tool/autoqa/lib/canonical-extractor.js
rayslxu: Yeah, the extractor will be updated soon with an actual parser. I will do a separate PR to clean this up altogether.

tool/autoqa/lib/bert-canonical-annotator.py
rayslxu: I searched around and didn't see a better solution... Any suggestion?

gcampax: We make our own vocabulary file (perhaps with wordnet as the starting point) and just load that. A single vocabulary file will be a lot smaller than nltk + wordnet.

tool/autoqa/auto-annotate.js
rayslxu: The current default is nothing - we skip auto annotation, but yeah, I think we should set it to the one that we think is the best

starter/schemaorg/Makefile
rayslxu: you can kill the other one

gcampax: What i mean is overwrite the old model and keep the old URL, but point to the new model.

rayslxu:
@gcampax I addressed most of the comments. 
For default options of algorithms, we will need a path to the paraphraser model if we want to set bart as the default, so I leave baseline as the default (skip auto annotate)
For the paraphraser model, I would rather keep the name.

PR #378, opened 2020-10-16
Wip/wikidata single turn

rayslxu:


tool/tokenize.js
```diff
@@ -0,0 +1,97 @@
...
+        process.nextTick(callback);
+    }
+}
+
+module.exports = {
+    initArgparse(subparsers) {
+        const parser = subparsers.add_parser('tokenize', {
+            add_help: true,
+            description: "Preprocess (tokenize) a string value dataset."
+        });
```
gcampax, line 73: This script is redundant with preprocess-string-dataset, right?

gcampax (approved):
Overall, looks good. Do we want to add the wikidata starter as a test in Travis? So it won't regress in the future.

tool/tokenize.js
rayslxu: oh, sorry. this was a script I used to re-tokenize training examples... it sneaked in by accident.

gcampax (approved):
Still yes

PR #379, opened 2020-10-20
Add templates for light-year, decade and century

s-jse:
This PR adds constant templates needed for three new measurement units added in https://github.com/stanford-oval/thingtalk-units/pull/30 PR.

### DIFF
--- a/languages/thingtalk/en/constants.genie
+++ b/languages/thingtalk/en/constants.genie
@@ -67,6 +67,8 @@ constant_date_point = {
             'the end of the week' => C.makeDate(new Ast.DateEdge('end_of', 'week'), '+', null);
             'the end of the month'=> C.makeDate(new Ast.DateEdge('end_of', 'mon'), '+', null);
             'the end of the year' => C.makeDate(new Ast.DateEdge('end_of', 'year'), '+', null);
+            'the end of the decade' => C.makeDate(new Ast.DateEdge('end_of', 'decade'), '+', null);
+            'the end of the century' => C.makeDate(new Ast.DateEdge('end_of', 'century'), '+', null);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


s-jse:
!turking is set for decade and century, but is not set for light-year. Does this make sense?

s-jse:
Also, last year/decade/century usually means "within the the last 1/10/100 years", not 
`'last year/decade/century'  => [C.makeDate(new Ast.DateEdge('start_of', 'year/decade/century'), '-', new Ast.Value.Measure(1, 'year')), C.makeDate(new Ast.DateEdge('start_of', 'year/decade/century')`

gcampax:
Not sure about your last comment. To me, "last year" means 2019, and "this year" means 2020. "Last year" is not the same as "in the last 12 months". "Last decade" and "last century" work the same: 2000-2010 and 1900-2000 resp.

Also, before merging this patch, you will need to:
- wait until a new release of thingtalk-units with the new units
- fix [date_utils](https://github.com/stanford-oval/thingtalk/blob/master/lib/date_utils.js) in thingtalk to support the new units

Also, what's the purpose for these units, other than "they are useful"? Is there a project that depends on these?

s-jse:
> 
> 
> Not sure about your last comment. To me, "last year" means 2019, and "this year" means 2020. "Last year" is not the same as "in the last 12 months". "Last decade" and "last century" work the same: 2000-2010 and 1900-2000 resp.
> 
> Also, before merging this patch, you will need to:
> 
>     * wait until a new release of thingtalk-units with the new units
> 
>     * fix [date_utils](https://github.com/stanford-oval/thingtalk/blob/master/lib/date_utils.js) in thingtalk to support the new units
> 
> 
> Also, what's the purpose for these units, other than "they are useful"? Is there a project that depends on these?

- I would say it is at least ambiguous (with two meanings). e.g. "Canada has enjoyed a booming economy in the last decade." means the last 10 years.

- These units are what I encountered while working on Schema2QA for Wikidata. For the Astronomy and People domains.

gcampax (commented):
Do we still care about this? Let's rebase/merge if so?

PR #383, opened 2020-10-27
Last preparation PR for TypeScript

gcampax:
Hopefully, the next PR will switch to ThingTalk 2.0
This is the last PR that prepares before that, we just add a ton of type annotations everywhere so all breakage will be caught right away.


rayslxu (approved):

gcampax:
Thanks for the review, I'm actually going to push a few more commits on top of this PR. All of it adding type annotations.

PR #384, opened 2020-10-28
sentence-generator: introduce a new, faster sampling strategy

gcampax:
Our previous sampling strategy consisted of enumerating, for each template, all possible expansions of that template, and then
sampling a portion of those expansions to call the semantic function with. This is not efficient. Instead, we can sample directly one candidate expansion without enumerating, if we are careful with setting up the probabilities correctly.

It's worth noting that previously we were doing _sampling without replacement_. The new algorithm samples _with replacement_ for speed. Practically, I don't think it makes a difference in terms of the quality of the generated datasets.

In dialogues, this gave me a 40% increase in generation throughput (turns/sec) with a small pruning size. I expect with a large pruning size it will yield even higher gains.

### DIFF
--- a/lib/dataset-tools/augmentation/replace_parameters.js
+++ b/lib/dataset-tools/augmentation/replace_parameters.js
@@ -329,7 +329,7 @@ export default class ParameterReplacer {
 
     _transformValue(sentenceValue, programValue, arg) {
         if (this._requotable)
-            return {sentenceValue, programValue};
+            return { sentenceValue, programValue };
 
         if (arg && arg.metadata.pluralize && coin(0.5, this._rng)) {
             const plural = this._paramLangPack.pluralize(sentenceValue);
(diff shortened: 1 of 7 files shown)


gcampax:
Actually, this fails full size generation:
```
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
assert_1.default(from.length > 0)
at expandRuleSample (/opt/genie-toolkit/dist/lib/sentence-generator/generator.js:1237:33)
at expandRule (/opt/genie-toolkit/dist/lib/sentence-generator/generator.js:1319:42)
at SentenceGenerator.generate (/opt/genie-toolkit/dist/lib/sentence-generator/generator.js:800:25)
at MinibatchDialogueGenerator._generateUser (/opt/genie-toolkit/dist/lib/sentence-generator/batch.js:219:29)
at MinibatchDialogueGenerator.nextTurn (/opt/genie-toolkit/dist/lib/sentence-generator/batch.js:284:14)
at DialogueGenerator._generateMinibatch (/opt/genie-toolkit/dist/lib/sentence-generator/batch.js:391:33)
```
I need to investigate why.

gcampax:
Another bit of measurement: on the MultiWOZ experiment, this cuts the time to generate the dataset from 10h to 5h. So it's a big win!

rayslxu (approved):

PR #386, opened 2020-11-06
Add filters for implicit identity and reverse property where table na…

rayslxu:
…me is absent

### DIFF
--- a/languages/thingtalk/load-thingpedia.ts
+++ b/languages/thingtalk/load-thingpedia.ts
@@ -995,6 +995,55 @@ export class ThingpediaLoader {
             span,
             {}
         ));
+
+        for (const arg of q.iterateArguments()) {
+            let op = '==';
+            let vtype : Type[] = [arg.type];
+            const slotOperator = arg.getImplementationAnnotation<string>('slot_operator');
+            if (slotOperator) {
+                op = slotOperator;
+            } else {
+                if (arg.type instanceof Type.Array) {
+                    vtype = [arg.type.elem as Type];
+                    op = 'contains';
+                } else if (arg.type.isRecurrentTimeSpecification) {
+                    vtype = [Type.Date, Type.Time];
+                    op = 'contains';
+                }
+            }
+
+            for (const type of vtype) {
+                const args : { [key : string] : Type } = {};
+                args[`p_${arg.name}`] = type;
+                const filter = new Ast.BooleanExpression.Atom(null, arg.name, op, new Ast.Value.VarRef(`p_${arg.name}`));
+                for (let canonical of arg.metadata.canonical.reverse_property || arg.metadata.canonical.npi || []) {
+                    if (!canonical.includes('#'))
+                        canonical += ' #';
+                    await this._loadTemplate(new Ast.Example(
+                        null,
+                        -1,
+                        'query',
+                        args,
+                        new Ast.Table.Filter(null, table, filter, table.schema),
+                        [canonical.replace('#', `\${p_${arg.name}:no-undefined}`)],
+                        [canonical.replace('#', `\${p_${arg.name}:no-undefined}`)],
+                        {}
+                    ));
+                }
+                if ('implicit_identity' in arg.metadata.canonical || ANNOTATION_RENAME.implicit_identity in arg.metadata.canonical) {
+                    await this._loadTemplate(new Ast.Example(
+                        null,
+                        -1,
+                        'query',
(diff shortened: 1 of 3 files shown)


languages/thingtalk/load-thingpedia.ts
```diff
@@ -995,6 +995,55 @@ export class ThingpediaLoader {
...
+                op = slotOperator;
+            } else {
+                if (arg.type instanceof Type.Array) {
+                    vtype = [arg.type.elem as Type];
+                    op = 'contains';
+                } else if (arg.type.isRecurrentTimeSpecification) {
+                    vtype = [Type.Date, Type.Time];
+                    op = 'contains';
+                }
+            }
```
gcampax, line 1013: You need to adjust `op` to be `=~` for String because you're not going through `makeFilter`

languages/thingtalk/load-thingpedia.ts
```diff
@@ -995,6 +995,55 @@ export class ThingpediaLoader {
...
+                    await this._loadTemplate(new Ast.Example(
+                        null,
+                        -1,
+                        'query',
+                        args,
+                        new Ast.Table.Filter(null, table, filter, table.schema),
+                        [canonical.replace('#', `\${p_${arg.name}:no-undefined}`)],
+                        [canonical.replace('#', `\${p_${arg.name}:no-undefined}`)],
+                        {}
+                    ));
```
gcampax, line 1031: You should check that the argument is input or output, and generate different code accordingly.

languages/thingtalk/load-thingpedia.ts
```diff
@@ -995,6 +995,55 @@ export class ThingpediaLoader {
...
+                    vtype = [Type.Date, Type.Time];
+                    op = 'contains';
+                }
+            }
+
+            for (const type of vtype) {
+                const args : { [key : string] : Type } = {};
+                args[`p_${arg.name}`] = type;
+                const filter = new Ast.BooleanExpression.Atom(null, arg.name, op, new Ast.Value.VarRef(`p_${arg.name}`));
+                for (let canonical of arg.metadata.canonical.reverse_property || arg.metadata.canonical.npi || []) {
```
gcampax, line 1019: You need to check if `metadata` is an object and not a string or array. I'm actually surprised this typechecks, `metadata.canonical` should be of type `unknown`.

languages/thingtalk/load-thingpedia.ts
```diff
@@ -995,6 +995,55 @@ export class ThingpediaLoader {
...
+                    vtype = [arg.type.elem as Type];
+                    op = 'contains';
+                } else if (arg.type.isRecurrentTimeSpecification) {
+                    vtype = [Type.Date, Type.Time];
+                    op = 'contains';
+                }
+            }
+
+            for (const type of vtype) {
+                const args : { [key : string] : Type } = {};
```
gcampax, line 1016: In the most recents PRs, I've started using `Record<string, Type>` instead of `{ [key : string] : Type }`. There is no lint rule to enforce that, but I think it's a bit more readable. They are equivalent otherwise.

languages/thingtalk/load-thingpedia.ts
```diff
@@ -995,6 +995,55 @@ export class ThingpediaLoader {
...
+                        -1,
+                        'query',
+                        args,
+                        new Ast.Table.Filter(null, table, filter, table.schema),
+                        [canonical.replace('#', `\${p_${arg.name}:no-undefined}`)],
+                        [canonical.replace('#', `\${p_${arg.name}:no-undefined}`)],
+                        {}
+                    ));
+                }
+                if ('implicit_identity' in arg.metadata.canonical || ANNOTATION_RENAME.implicit_identity in arg.metadata.canonical) {
```
gcampax, line 1033: I think at some point you mentioned that `implicit_identity` is redundant with `reverse_property`. Do we still need it then?

gcampax (commented):
I think this looks good, probably worth adding some tests by adding some `implicit_identity` or `reverse_property` to multiwoz (attraction-type should fit the bill). Some comments too.

languages/thingtalk/load-thingpedia.ts
rayslxu: yes, `implicit_identity` is just for compatibility at this point.

gcampax (approved):
Is this ready? I think it looks good to merge to `next` if tested.

PR #387, opened 2020-11-06
Fix running the dialogue agent after building

gcampax:
If genie is run from the dist/ directory, the relative path to
the compiled templates is different.

Tests did not catch this because they use ts-node and run genie
from the source directory.


gcampax:
@s-jse this should fix the problem you were having with `genie assistant`

rayslxu (approved):

PR #388, opened 2020-11-09
Look at dialog context as well when detecting slots

s-jse:


gcampax (changes requested):
Yeah this should works for dialogues, and should basically work for the user side. In the code you need to be careful though, for single sentence you won't have a context at all and the whole thing will break. You should not run this code if there is no context. I'm surprised it typechecks in fact.

s-jse:
> 
> 
> Yeah this should works for dialogues, and should basically work for the user side. In the code you need to be careful though, for single sentence you won't have a context at all and the whole thing will break. You should not run this code if there is no context. I'm surprised it typechecks in fact.

I added the check.

lib/dataset-tools/augmentation/replace_parameters.ts
```diff
@@ -678,7 +671,47 @@ export default class ParameterReplacer {
...
         }
 
-        for (const slot of program.iterateSlots2()) {
+        const entities : ThingTalk.NNSyntax.EntityMap = {};
+        // replace all entities with SLOT_*, which allows us to pass a VarRef instead of a real value
+        const replaced_code = this._replaceWithSlot(code, entities);
+        const target_program = ThingTalk.NNSyntax.fromNN(replaced_code, entities);
+        await target_program.typecheck(this._schemas, true);
+        
+        if (context != null && context.length > 0) {
```
gcampax, line 680: Don't use `!=`, always use `!==`. Also, the context will not be null, undefined, or the empty string. It will be the empty array (see type definition), so you need use `context.length !== 0`.

Also, please run `npx make` and `npm run lint` to find all these problems.
Normally the PR would be blocked because Travis would complain, but Travis is down at the moment...

gcampax (approved):
Please fix, and also run the tests. Good to merge then.

PR #393, opened 2020-11-30
NAACL submission work

gcampax:
Assorted patches accumulated in the last month


gcampax:
Tests don't pass because the batching code needs a newer version of genienlp, I'll ask @s-jse to submit a PR for it

PR #398, opened 2020-12-04
Port to ThingTalk 2.0

gcampax:
It finally happened: here is the PR that moves all of Genie to ThingTalk 2.0
It's a big PR, but surprising not as big as I thought it would be, considering there is very little legacy code left and the result is relatively clean despite all compatibility paths.


gcampax:
Note: many commits are from #393. I will rebase this PR once the other one lands.

rayslxu (approved):

PR #399, opened 2020-12-07
Port schema.org starter to ThingTalk 2.0

gcampax:
Mostly so @s-jse can see the new syntax.

Depends on https://github.com/stanford-oval/thingtalk/pull/281 if you want to typecheck the datasets again, otherwise dates will fail to be recognized.


s-jse (approved):
Thanks. The new surface syntax definitely looks a lot more suitable for pre-trained language models.

PR #402, opened 2020-12-13
PosParser for annotation extraction from paraphrase

rayslxu:
A parser based on part-of-speech information to extract canonical annotations from paraphrases. 
It's implemented as a regular language using NFA, with support for the following operators:
- `*` closure
- `_` concat (implicit)
- `|` union
- `()` parentheses 
- `.` wildcard 

In addition, it uses `[]` to indicate matching group. 

This generates exactly the same annotations as the old code for the 6 schema.org domains.

### DIFF
--- a/lib/pos-parser/nfa.ts
+++ b/lib/pos-parser/nfa.ts
@@ -103,9 +103,14 @@ export class NFA {
     }
 
     private preprocess(utterance : string, domainCanonicals : string[], value : string) : Token[] {
+        // remove punctuation at the end
+        utterance = utterance.replace(/[.,?!;]\s*$/g, '');
+
         const tokenized : TokenizerResult = this.tokenizer.tokenize(utterance);
         const tokens = tokenized.rawTokens;
         const posTags : Array<string|undefined> = this.languagePack.posTag(tokenized.rawTokens);
+
+
 
         // replace value with special token $value
         if (arrayMatchCount(tokens, value.split(' ')) !== 1)
@@ -116,10 +121,17 @@ export class NFA {
         tokens[valueIndex] = '$value';
         tokens.splice(valueIndex, value.split(' ').length - 1);
 
+        // expand domain canonicals to include plurals
+        const domainCanonicalsExpanded : Set<string> = new Set();
+        for (const canonical of domainCanonicals) {
+            domainCanonicalsExpanded.add(canonical);
+            domainCanonicalsExpanded.add(this.languagePack.pluralize(canonical));
+        }
+
         // replace domain canonical with special token $domain
         let domainCanonical : string;
         let matchCount = 0;
-        for (const canonical of domainCanonicals) {
+        for (const canonical of domainCanonicalsExpanded) {
             const count : number = arrayMatchCount(tokens, canonical.split(' '));
             if (count > 0)
                 domainCanonical = canonical;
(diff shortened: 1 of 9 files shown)


lib/pos-parser/nfa.ts
```diff
@@ -43,27 +47,113 @@ class State {
...
+            if (a[i + j] !== b[j])
+                match = false;
+        }
+        if (match)
+            count += 1;
+    }
+    return count;
+}
+
 export class NFA {
```
gcampax, line 112: At this point, NFA is not just a generic NFA engine, it's specifically for your canonical form parser. Maybe split this into two classes, one that just implements the NFA and another that has the preprocessing and POS logic?

test/unit/test_infix_to_postfix.js
```diff
@@ -22,12 +22,19 @@ import assert from 'assert';
...
     ['( a * | b ) c', 'a * b | c _'],
     ['a b *', 'a b * _'],
-    ['a . *', 'a . * _']
+
+    // wild card
+    ['a . *', 'a . * _'],
+
+    // capturing group
+    ['a [ b ]', 'a [ b ] _'],
+    ['a [ . * c ]', 'a [ . * _ c ] _']
```
gcampax: Using [ ] for capturing groups is weird given the normal semantics of regular expressions...

test/unit/test_pos_nfa.js
```diff
@@ -23,18 +23,20 @@ import { toNFA } from '../../lib/pos-parser/nfa';
 
 const TEST_CASES = [
     [
-        '( show me | find me | find | search for ) a [ $value ] $domain', ['restaurant', 'diner'], 'chinese',
+        '( show me | find me | find | search for ) ( a | ε ) [ $value ] $domain', ['restaurant', 'diner'], 'chinese',
```
gcampax: Maybe you should add the `?` operator as syntactic sugar in your regex DSL

lib/pos-parser/nfa.ts
```diff
@@ -84,6 +84,31 @@ function arrayMatchCount(a : string[], b : string[]) : number {
...
+
+    if (current.pos === 'IN' && current.token === 'with')
+        return false;
+
+    if (current.pos.startsWith('V')) {
+        if (['has', 'have', 'had'].includes(current.token))
+            return next && ['VBN', 'VBG'].includes(next.pos!);
+        if (['is', 'are', 'was', 'were'].includes(current.token))
+            return false;
+    }
```
gcampax, line 107: I wonder if you should instead have multiple templates match the sentence, and use a scoring function to determine which wins...

gcampax (commented):
I am honestly impressed that you implemented an NFA-based regular expression engine! I would have used backtracking and be done with it...
The code is very nice, clean and readable, which quite something for a non-trivial algorithm.

I have a few comments, but overall, let's merge this!

test/unit/test_pos_nfa.js
rayslxu: > I am honestly impressed that you implemented an NFA-based regular expression engine! I would have used backtracking and be done with it...
> The code is very nice, clean and readable, which quite something for a non-trivial algorithm.
> 
> I have a few comments, but overall, let's merge this!

Thanks! The comments are all good - but let's merge this and deal with them later (once tests pass)

Issue #252, opened 2020-07-29
Identically named parameters should not be reused across devices


gcampax:
Because they are likely to have different canonical forms, with different meaning.
Case in point: we currently generate "hotels with train number carolina bed and breakfast" because "train number" is a canonical form of "id"...

gcampax:
I originally planned to postpone this until after Almond 1.99, because I thought it would affect the dataset for training only, but actually it affects the agent reply at inference time (because of course, the agent uses the agent templates to formulate the reply) so we need to fix it this week. Adjusting the milestone.

gcampax:
The most pressing part of this issue was fixed. I'm going to remove this from the milestone.

gcampax:
Fixed by #418

PR #418, opened 2021-01-06
[AL1-4] Overhaul synthesis

gcampax:
This PR started as a simple feature: the ability to confirm any command (not just simple invocations of actions), using user-specified strings, removing `#[confirmation]`, and efficiently.

It snowballed into a huge PR that refactors synthesis to be a lot more efficient. The reason is that otherwise exhaustively generated bottom-up confirmations just don't work, the agent never generates the right sentence because it generates too much irrelevant stuff. This is a problem at training time (sampling issue) and at inference time (too slow).

Hence, I decided to introduce efficient hybrid top-down/bottom-up generation.
The refactor involves two ideas. First, separating all the natural language variations out, and handling those purely top-down, using string-interp/ICU MessageFormat like format strings in place of templates. We need those for I18n (especially with highly inflected languages) for string customizability (that Monica insists on), and to generate grammatically correct English on the agent side (ie handle singular/plural correctly).
The second idea is to make semantic functions a lot more efficient through hashtable indices, such that most of the pruning is done by the table, and the semantic function only needs to check for complex logic. This allows us to match contexts efficiently, so we can basically make every piece of agent sentence into a context-dependent non-terminal (which otherwise would lead to terrible performance). And it allows us to match functions and parameters, hence fixing long-standing issue #252 

The PR is not finished yet. I will edit this description when I'm done. But you can start reading the details of the new ideas in the existing commit messages.


gcampax:
I'm not done with synthesis changes (I still want to refactor how we specify the natural language bits of the templates, and how we construct sentences at the end), but I think this is a good place to start reviewing. @rayslxu please take a look, very carefully because it's a tricky PR, despite tests and type-checking.

gcampax:
Timing and dataset quality statistics are at https://docs.google.com/spreadsheets/d/159oZV2aE4Jy7lTzyweSIuy03tU4rh1HPPIr8M8A63wo/edit#gid=710122031
For a small dataset on the current version of thingpedia-common-devices, we get a 2x speedup with the whole PR.

gcampax:
At first, I thought the accuracy issue was a dataset balancing problem, because the balance between old synthesis and new synthesis is quite different (new synthesis is a lot more uniform, which means a lot of tail commands and fewer important commands), so I wrote https://github.com/stanford-oval/genie-toolkit/pull/418/commits/201940d1aba0d9fbe13a526ddff07ba956bc05ec . After finding https://github.com/stanford-oval/genie-toolkit/pull/418/commits/35b437613ef51490ff707a370a3c3a61b995122b which was an actual bug that manifests with zero accuracy on all genre questions in Spotify, I am not sure that's necessary. But if we keep the tuning change, we can push the pruning size to 500 (up to 200 or 250 before) which gives a lot of interesting things.

rayslxu (approved):
It looks great! Given the speed boost and performance we get, I focused more on the design and understanding the flow rather than checking every implementation detail. 
I'm not a huge fan of the `with` syntax as it is a little confusing especially if the non terminal is in the middle of the template. But I could not think of a better way. I also saw your design note for the second part of Genie refactoring, it would solve this confusion if we put the `with` in the parameter declaration part. (well, it will no longer be valid TS syntax as you hope...)

gcampax:
Yeah part of the reason for the second part of refactoring is that with "with" the syntax is currently messy and unreadable. I though about various ways to make it valid-ish TS syntax. The best I could come up with is:
```
$root = {
    ["${foo}"]: (foo : foo_bar[{prop : true}]) => C.makeRoot(foo),
    
    ["${a} ${b}"]: (a : foo_bar, b : foo_bar[{ prop : a.prop }]) => C.makeRoot(a, b),
};
```
This is valid TypeScript (syntactically, not semantically), to the extend that I don't get TSLint errors. But it's also very ugly.
What do you think?

rayslxu:
> Yeah part of the reason for the second part of refactoring is that with "with" the syntax is currently messy and unreadable. I though about various ways to make it valid-ish TS syntax. The best I could come up with is:
> 
> ```
> $root = {
>     ["${foo}"]: (foo : foo_bar[{prop : true}]) => C.makeRoot(foo),
>     
>     ["${a} ${b}"]: (a : foo_bar, b : foo_bar[{ prop : a.prop }]) => C.makeRoot(a, b),
> };
> ```
> 
> This is valid TypeScript (syntactically, not semantically), to the extend that I don't get TSLint errors. But it's also very ugly.
> What do you think?

Why brackets? It could be parathesis (like a function) or `<>` (like template); semantically they makes (slightly) more sense....

gcampax:
> Why brackets? It could be paratheses (like a function) or `<>` (like template); semantically they make (slightly) more sense....

`:` or `=` are not allowed in angle brackets, and parenthesis don't work. If we give up on valid TypeScript, then yeah my vote would be for:
```typescript
$root = {
    "${foo}": (foo : foo_bar<prop=true>) => C.makeRoot(foo),
    
    "${a} ${b}": (a : foo_bar, b : foo_bar<prop=a.prop>) => C.makeRoot(a, b),
};
```

PR #420, opened 2021-01-11
Add a new conversation API to see the raw ThingTalk results

gcampax:
Every time a new ThingTalk program is executed, send a "new-program"
message on the conversation web socket with the unique ID of the
program (in case the program is long running), the executed code,
and the raw results returned by the Thingpedia function.

This should allow a consumer of the conversation API to display
the results of the program in a custom graphical way (eg a chart)

This PR is needed for the finance assistant project.


rayslxu (approved):

gcampax:
This is not critical, so I'm going to hold this until after the Alpha release.

PR #422, opened 2021-01-14
Synthesis overhaul, the uncontroversial bits

gcampax:
PR #418 is getting long and it's far from finished. So I'm sending this PR first, so we can start merging that one piecewise instead of waiting, in case that one goes nowhere.


rayslxu (approved):
looks good

PR #424, opened 2021-01-25
Add kf-serving inference backend

jgd5:
The inference service will be accessible within cluster using following HTTP.
We use the model/lang as inference service name, but the name is limited to alphanumeric.  

For example, the name for `@org.thingpedia.models.contextual/en` is `orgthingpediamodelscontextualen` .

The service can be accessed with following HTTP call:

URL:  http://cluster-local-gateway.istio-system.svc.cluster.local/v1/models/orgthingpediamodelscontextualen:predict
HTTP Header {Host: orgthingpediamodelscontextualen.kfserving-test.svc.cluster.local }

### DIFF
--- a/lib/prediction/localparserclient.ts
+++ b/lib/prediction/localparserclient.ts
@@ -45,8 +45,8 @@ const NLG_QUESTION = 'what should the agent say ?';
 export interface LocalParserOptions {
     id ?: string;
     nprocesses ?: number;
-    kf_inference_ingress ?: string;
-    kf_inference_domain ?: string;
+    kfInferenceIngress ?: string;
+    kfInferenceDomain ?: string;
 }
 
 function compareScore(a : PredictionCandidate, b : PredictionCandidate) : number {
--- a/lib/prediction/localparserclient.ts
+++ b/lib/prediction/localparserclient.ts
@@ -45,6 +45,8 @@ const NLG_QUESTION = 'what should the agent say ?';
 export interface LocalParserOptions {
     id ?: string;
     nprocesses ?: number;
+    kf_inference_ingress ?: string;
+    kf_inference_domain ?: string;
 }
 
 function compareScore(a : PredictionCandidate, b : PredictionCandidate) : number {
@@ -75,7 +77,7 @@ export default class LocalParserClient {
         this._locale = locale;
         this._langPack = I18n.get(locale);
         this._tokenizer = this._langPack.getTokenizer();
-        this._predictor = new Predictor(options.id || 'local', modeldir, options.nprocesses);
+        this._predictor = new Predictor(modeldir, options);
 
         this._exactmatcher = exactmatcher;
         this._tpClient = tpClient;
--- a/lib/prediction/predictor.ts
+++ b/lib/prediction/predictor.ts
@@ -26,6 +26,8 @@ import * as child_process from 'child_process';
 import JsonDatagramSocket from '../utils/json_datagram_socket';
 import HttpEmitter from '../utils/http_emitter';
 
+import { LocalParserOptions } from './localparserclient';
+
 const DEFAULT_QUESTION = 'translate from english to thingtalk';
 
 interface PredictionCandidate {
@@ -52,7 +54,9 @@ interface Example {
 
 class Worker extends events.EventEmitter {
     id : string;
-    private _inference_name : string;
+    private _kf_inference_name : string;
+    private _kf_inference_ingress ?: string;
+    private _kf_inference_domain ?: string;
     private _child : child_process.ChildProcess|null;
     private _hadError : boolean;
     private _stream : JsonDatagramSocket|HttpEmitter|null;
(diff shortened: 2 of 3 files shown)


lib/prediction/predictor.ts
```diff
@@ -282,7 +303,8 @@ export default class Predictor {
     }
 
     private _startWorker() {
-        const worker = new Worker(`${this.id}/${this._nextId++}`, this._modeldir);
+        const inference_name = this.id.replace(/\W/g, '');
```
gcampax, line 306: Do we need some escaping mechanism instead of dropping characters? To some extent, model IDs can be user-generated, and I worry that malicious users can craft model IDs that give them access to models they do not own.

lib/utils/http_emitter.ts
```diff
@@ -0,0 +1,70 @@
...
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
```
gcampax, line 39: Duplicate copyright comment?

lib/utils/http_emitter.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// HttpEmitter wraps http calls in EventEmitter interface
+export default class HttpEmitter extends events.EventEmitter {
+    private _url : string;
+    private _host : string;
+
+    constructor(url : string, host : string) {
+        super();
+        this._url = url;
+	this._host = host;
```
gcampax, line 52: Bad indent, please use 4 spaces everywhere.

lib/prediction/predictor.ts
```diff
@@ -90,34 +94,46 @@ class Worker extends events.EventEmitter {
...
-            //console.error(`Child exited with code ${code}, signal ${signal}`);
-
-            this._failAll(new Error(`Worker died`));
-            this._child = null;
-            this.emit('exit');
-        });
-
-        this._stream = new JsonDatagramSocket(this._child.stdout!, this._child.stdin!, 'utf8');
-
+        if (process.env.USE_KF_SERVING_INFERENCE) {
```
gcampax, line 97: We should probably use a parameter to Predictor or Worker instead of an environment variable, especially one that is not properly namespaced. I imagine almond-cloud will set the parameter if the almond-cloud config file says so, and other environments (eg eval step in kubeflow) will not set the parameter all.

gcampax (commented):
It's a bit messy isn't it? Can we refactor out the parts of Predictor that have to do with the JSON protocol (and promise/request wrangling) from the minibatch handling code, so we don't need to make this confusing HTTP request wrapper?

lib/prediction/predictor.ts
jgd5: It was unfortunate kfserving service name can only be alphanumeric.  This name is also hard coded in ingress routing rule.
I'm open to ideas. We can add a config map of model name to inference service name.  But maintaining it would be painful.   I suppose we will have  access control checks before the prediction? We can also force model names to be alphanumeric only.

jgd5:
> It's a bit messy isn't it? Can we refactor out the parts of Predictor that have to do with the JSON protocol (and promise/request wrangling) from the minibatch handling code, so we don't need to make this confusing HTTP request wrapper?

I'm still learning the idiomatic ways in JS/TS.  I'm not seeing an easy way to refactor this.  Maybe I'll ping you on slack.   If it's too much change we can break it out in another PR or do it after the migration.

lib/prediction/predictor.ts
gcampax: Do they allow any special character? -, _, anything? If they do, we can do quoted-printable escaping and replacing all special characters with `_` followed by the hex representation of the character.
I tried looking up the docs but didn't find anything definitive, although I did find examples with _ and -

lib/prediction/localparserclient.ts
```diff
@@ -45,6 +45,8 @@ const NLG_QUESTION = 'what should the agent say ?';
 export interface LocalParserOptions {
     id ?: string;
     nprocesses ?: number;
+    kf_inference_ingress ?: string;
+    kf_inference_domain ?: string;
```
gcampax, line 49: Not a huge deal, but we try to use lowerCamelCase in JS.

gcampax (approved):
If you want to postpone the refactoring, that's fine. Please fix the indent, you have quite a few stray tabs. Other than that I think it's good to merge.

lib/prediction/predictor.ts
jgd5: Kubenetes name allows '- and .  but I did not have success using them.  The inference service will not create resources after I applied.  I didn't see any error message but I haven't had the time find the root cause. I suspect one of resources failed to create due to the name.

jgd5: We can also use 0 as special character. So any non-alphanumeric can be represented by 0??  (where ?? is the ASCII hex code).  If 0 exists in the name, we can escape it using 00?

gcampax: Yeah 0 as special character would work.

jgd5: cool.. will do that in another PR.

PR #425, opened 2021-01-27
Add two commands to run the agent in simulation mode

gcampax:
This PR adds two commands:

- `genie interactive-annotate` replaces the old `demo-dialog`: it runs an interactive command-line assistant, it takes commands, lets the user annotate those commands, then executes or simulates the commands and runs the agent; the output is written to a dataset
- `genie simulate-dialogs` takes a dialogue dataset and runs the agent on the last turn of each dialogue (according to the user target annotation in the dataset), extending the dialogue by one half-turn 

The first command should be helpful to annotate Mario commands. It is a better option than `manual-annotate-dialog` because there is no WoZ data to start from.
The second command should be useful to Sina for the error detection project, as we can create a dataset with a bunch of errors and run this command to see how the agent would surface those errors to the user.


s-jse (approved):

s-jse:
A useful option for `simulate-dialogs` would be something like `use_gold_policy` where it reads the policy from input instead of coming up with it. It is useful for when we want to compare the agent response with gold independent of policy decisions.

gcampax:
I'm not sure that is implementable. The policy constructs the utterance and the target state at the same time, by searching through the templates. You can't have the policy construct a specific state.

Besides a lot of agent states in say MultiWOZ are unreachable at inference time, because at inference time we explicitly restrict the set of agent actions to minimize user confusion.
If you need that for sentence generation (instead of error correction), your best bet is to search among the training set for a compatible context and find if by chance we have generated the right agent state. There aren't that many agent states in MultiWOZ so this should be feasible.

PR #426, opened 2021-01-28
Refactor prediction code for kubeflow

gcampax:
We don't actually need three separate options for Kubeflow,
all we need is one URL. We can recognize it with a pseudo-scheme
"kf+http(s):" (to distinguish from the RemoteParserClient) and
let the user specify it normally.

Also, refactor Predictor to separate the batching code from
the request code. Remove the ability to have multiple parallel
workers, as multiple workers on the same machine is not very
useful (among other things, they all run on the same GPU), and
move the batching code up. Refactor the Kubeflow backend to be
a straight-forward HTTP call instead of wrapping into a stream.

@jgd5 this is what I had in mind as refactor, although it's a bit more aggressive because I also removed the multi-worker code given that KFServing will take care of load-balancing.

Tested with `genie assistant --nlu-server 'kf+http://nlp.kfserving-test.svc.cluster.local/v1/models/nlp:predict'` from a Jupyter notebook. I'll test batch evaluation too before merging.


lib/prediction/predictor.ts
```diff
@@ -52,92 +49,61 @@ interface Example {
...
-    private _kfInferenceIngress ?: string;
-    private _kfInferenceDomain ?: string;
+class LocalWorker extends events.EventEmitter {
     private _child : child_process.ChildProcess|null;
-    private _hadError : boolean;
-    private _stream : JsonDatagramSocket|HttpEmitter|null;
+    private _stream : JsonDatagramSocket|null;
     private _nextId : 0;
     private _requests : Map<number, Request>;
     private _modeldir : string;
```
jgd5, line 57: Nit: maybe rename this to _modelPath since it can be url or file path?

lib/prediction/predictor.ts
```diff
@@ -224,7 +226,7 @@ class Worker extends events.EventEmitter {
         this._minibatchStartTime = now;
```
jgd5, line 226: can `assert(this._minibatch.length === 0)` be false when `task !=  this._minibatchTask`?

lib/prediction/predictor.ts
```diff
@@ -160,26 +126,81 @@ class Worker extends events.EventEmitter {
...
+    stop() {}
+
+    async request(task : string, minibatch : Example[]) : Promise<PredictionCandidate[][]> {
+        const response = await Tp.Helpers.Http.post(this._url, JSON.stringify({
+            id: 0, // should be ignored
+            task,
+            instances: minibatch
+        }), { dataContentType: 'application/json', accept: 'application/json' });
+        const parsed = JSON.parse(response);
+        return JSON.parse(parsed.predictions).instances.map((instance : any) : PredictionCandidate[] => {
```
jgd5, line 173: Maybe add a TODO  to remove this extra  JSON.parse() after genienlp kfserver is updated?

lib/prediction/predictor.ts
```diff
@@ -52,92 +49,61 @@ interface Example {
     reject(err : Error) : void;
```
jgd5, line 49:
```
const MINIBATCH_SIZE = 30;
const MAX_LATENCY = 50; //
````
Maybe make these two into parameters? Could be useful when we starting tuning GPU serving.

jgd5 (approved):
The refactoring looks great. It's much cleaner. 
All minor comments and one question about the assert.

lib/prediction/predictor.ts
gcampax: We always flush the request before starting a new request, so the minibatch will always be empty.

lib/prediction/predictor.ts
gcampax: Good idea

PR #428, opened 2021-01-30
[AL1-13] Synthesis overhaul, part 2 and a half

gcampax:
(Part 1 of the synthesis overhaul was #422 , part 2 was #418 , and part 3 will be the [new template design](https://wiki.almond.stanford.edu/en/genie/design-notes/new-template-language))

This PR introduces a number of changes in how we construct phrases for confirmations and action results, finally solving the ugly "Your request was completed successfully" reply for Spotify, and allowing us to confirm any ThingTalk program, no matter how complicated. The latter is a prereq to supporting streams again. We also change how we construct phrases for query results, and finally the developer wishes in Thingpedia are respected entirely, with no randomness.

As an aside, I realized that the use of a singleton for the ThingpediaLoader class will cause enormous issues when users share the same process in almond-cloud (or when the same user opens two tabs, also in almond-cloud), so the last commit removes that singleton. The code in the end is ugly, but it's _more_ correct.


lib/utils/thingtalk/describe.ts
```diff
@@ -549,58 +569,226 @@ export class Describer {
...
         for (const attr of selector.attributes) {
             if (attr.name === name)
                 return this._describeArg(attr.value, {});
         }
         return undefined;
     }
 
+    private _computeParamMatchingScore(exampleInParams : Ast.InputParam[],
+                                       programInParams : Ast.InputParam[],
+                                       exampleArgs : Record<string, Type>) {
```
rayslxu, line 602: It's minor, but there are a few cases the function type is not annotated...

rayslxu (approved):
looks good. Just some minor format issues

lib/utils/thingtalk/describe.ts
gcampax: We don't require type annotations on function return types when it's inferrable, do we? Otherwise there are a lot of places where we need to add types.

PR #430, opened 2021-01-30
[AL1-12] Implement FAQ based meta-commands

gcampax:
I originally thought that the meta-skill needed the synthesis refactoring as a prereq, but it turns out that it doesn't. We can implement the basic FAQ commands of the meta-skill without other changes in Genie.
See the commit message for the explanation of the implementation.

As we discussed, the content is specified in a data file, faq.yaml.
@monicalam please edit that file going forward instead of the spreadsheet. The format should be straight-forward.
Note that questions should all include `?`. Synthesis automatically removes it for a subset of training data.

I also removed the "device" and "command" queries that look up the list of skills and commands in Thingpedia, because they would conflict with the "general help" FAQ, and just dumping Thingpedia is not very useful.


rayslxu (approved):
ha, nice trick with action + enum. lgtm.

PR #442, opened 2021-02-03
`simulate-dialogs` with parser

s-jse:
This PR makes the following changes to `genie simulate-dialog`:
- Adds the option to parse the last user utterance using a neural semantic parser, instead of reading UT from file.
- For each input dialogue, output `n` partial dialogues where `n` is the number of turns in that dialogue. The `i`-th output has the entire conversation from the beginning till turn `i`.
- Adds the option to only output partial dialogues where the provided semantic parser makes a mistake. i.e. if a user turn is parsed correctly, it will not be output.

### DIFF
--- a/tool/simulate-dialogs.ts
+++ b/tool/simulate-dialogs.ts
@@ -40,6 +40,7 @@ import * as ParserClient from '../lib/prediction/parserclient';
 
 import { readAllLines } from './lib/argutils';
 import MultiJSONDatabase from './lib/multi_json_database';
+import { PredictionResult } from '../lib/prediction/parserclient';
 
 export function initArgparse(subparsers : argparse.SubParser) {
     const parser = subparsers.add_parser('simulate-dialogs', {
@@ -81,6 +82,11 @@ export function initArgparse(subparsers : argparse.SubParser) {
         help: `The URL of the natural language server to parse user utterances. Use a file:// URL pointing to a model directory to use a local instance of genienlp.
                If provided, will be used to parse the last user utterance instead of reading the parse from input_file.`
     });
+    parser.add_argument('--output_mistakes_only', {
+        action: 'store_true',
+        help: 'If set and --nlu-server is provided, will only output partial dialogues where a parsing mistake happens.',
+        default: true
+    });
 }
 
 class SimulatorStream extends Stream.Transform {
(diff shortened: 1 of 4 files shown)


tool/simulate-dialogs.ts
```diff
@@ -129,14 +199,53 @@ class SimulatorStream extends Stream.Transform {
...
+}
+
+class DialogueToPartialDialoguesStream extends Stream.Transform {
+
+    constructor() {
+        super({ objectMode : true });
+    }
+
+    private copyDialogueTurns(turns : Array<DialogueTurn>) : Array<DialogueTurn> {
+        const copy = new Array<DialogueTurn>();
```
gcampax, line 224: `copy : DialogueTurn[] = []` is more readable and faster at runtime (due to JS interpreter shenanigans)

tool/simulate-dialogs.ts
```diff
@@ -129,14 +199,53 @@ class SimulatorStream extends Stream.Transform {
...
+    }
+}
+
+class DialogueToPartialDialoguesStream extends Stream.Transform {
+
+    constructor() {
+        super({ objectMode : true });
+    }
+
+    private copyDialogueTurns(turns : Array<DialogueTurn>) : Array<DialogueTurn> {
```
gcampax, line 223: By convention, private methods start with `_` (like in Python)

tool/simulate-dialogs.ts
```diff
@@ -169,11 +278,22 @@ export async function execute(args : any) {
...
+    if (args.nlu_server){
+        parser = ParserClient.get(args.nlu_server, args.locale);
+        await parser.start();
+    }
+
     await StreamUtils.waitFinish(
         readAllLines(args.input_file, '====')
         .pipe(new DialogueParser())
-        .pipe(new SimulatorStream(policy, simulator, schemas))
+        .pipe(new DialogueToPartialDialoguesStream())
```
gcampax, line 290: Do you want to do this transformation unconditionally?

tool/simulate-dialogs.ts
```diff
@@ -169,11 +278,22 @@ export async function execute(args : any) {
...
+        .pipe(new SimulatorStream(policy, simulator, schemas, parser, tpClient, args.output_mistakes_only, args.locale))
         .pipe(new DialogueSerializer())
         .pipe(args.output)
     );
+
+    if (parser !== null) {
+        await parser.stop();
```
gcampax, line 297: No braces around single line if bodies, unless the other branch needs braces (eslint will complain)

tool/simulate-dialogs.ts
```diff
@@ -75,36 +77,104 @@ export function initArgparse(subparsers : argparse.SubParser) {
...
     });
+    parser.add_argument('--nlu-server', {
+        required: false,
+        help: `The URL of the natural language server to parse user utterances. Use a file:// URL pointing to a model directory to use a local instance of genienlp.
+               If provided, will be used to parse the last user utterance instead of reading the parse from input_file.`
+    });
+    parser.add_argument('--output_mistakes_only', {
+        action: 'store_true',
+        help: 'If set and --nlu-server is provided, will only output partial dialogues where a parsing mistake happens.',
+        default: true
```
gcampax, line 88: If you set default: true, then you need the negated version of this option, otherwise there is no way to unset it.

gcampax (approved):
This looks good overall, a couple comments, and then yeah merge when tests pass.

s-jse:
Travis CI is stuck in queue for 12 hours now. @gcampax is this normal?

PR #443, opened 2021-02-03
Fix weather skill

gcampax:
The weather skill reveals a lot of bugs in how we handle questions to resolve the user's stored locations.


rayslxu (approved):
Yeah, this is straightforward. But why don't we have any location answers in training?

gcampax:
Those location answers are performed using the old `ask()` internal API, not as part of the regular state machine. There is no formal state for "the agent is resolving the meaning of home" - yet. I plan to do that eventually, but not for Alpha.

PR #448, opened 2021-02-04
Dialogue logging with test mode

rayslxu:


lib/dataset-tools/parsers.ts
```diff
@@ -206,14 +210,19 @@ class DialogueSerializer extends Stream.Transform {
...
                 if (this._annotations)
-                    this._pushMany(this._prefixLines(turn.agent_target!, 'AT: '));
+                    this._pushMany(this._prefixLines(turn.agent_target, 'AT: '));
                 if (this._annotations && turn.intermediate_context)
                     this._pushMany(this._prefixLines(turn.intermediate_context, 'C: '));
+
+                if (turn.rate)
+                    this._pushMany(this._prefixLines(turn.rate, '#! '));
+                if (turn.comment)
+                    this._pushMany(this._prefixLines(turn.comment, '# '));
```
gcampax, line 223: No comment at the first turn?

lib/dataset-tools/parsers.ts
```diff
@@ -266,7 +275,7 @@ class DialogueParser extends Stream.Transform {
             if (/^(R)?(P)?(C)?(S)?(E)?$/.test(this._id))
                 this._id = '_' + this._id;
         }
-        if (!line || line.startsWith('#')) {
+        if (!line) {
```
gcampax, line 278: So, the first comment line is parsed both as the ID of the dialogue and the comment for the first turn?
Also, how do comment work if they're not in the middle of a turn, they get assigned to whichever turn follows them?

lib/dialogue-agent/conversation.ts
```diff
@@ -548,4 +570,48 @@ export default class Conversation extends events.EventEmitter {
             console.log('Almond sends link: '+ url);
         return this._addMessage({ type: MessageType.LINK, url, title });
     }
+
+    appendLog(turn : DialogueTurn) {
+        this._log.push(turn);
+    }
```
gcampax, line 576: You'll need to create multiple dialogues, not just one, and move to the next dialogue every time the dialogue loop resets the state. Otherwise the state tracked at runtime and the state tracked in the file (by repeated applications of computeNewState) will be different.

lib/dialogue-agent/conversation.ts
```diff
@@ -548,4 +570,48 @@ export default class Conversation extends events.EventEmitter {
...
+    commentLast(comment : string) {
+        if (this._log.length === 0)
+            throw new Error('No response found to be commented');
+        const last = this._log[this._log.length - 1];
+        last.comment = comment;
+    }
+
+    async saveLog() {
+        const dir = this._engine.platform.getWritableDir();
+        const logfile = path.join(dir, this.id + '.txt');
```
gcampax, line 601: Should we make a subdirectory dedicated to conversation logs?

lib/dialogue-agent/conversation.ts
```diff
@@ -548,4 +570,48 @@ export default class Conversation extends events.EventEmitter {
...
+        serializer.pipe(output);
+        serializer.write({ id: this.id, turns : this._log });
+        serializer.end();
+
+        await StreamUtils.waitFinish(output);
+    }
+
+    loadLog() {
+        const dir = this._engine.platform.getWritableDir();
+        return fs.readFileSync(path.join(dir, this.id + '.txt')).toString();
```
gcampax, line 615: Is this for download? Maybe return a stream or filename instead of a blob?

lib/dialogue-agent/conversation.ts
```diff
@@ -548,4 +570,48 @@ export default class Conversation extends events.EventEmitter {
             console.log('Almond sends link: '+ url);
         return this._addMessage({ type: MessageType.LINK, url, title });
     }
+
+    appendLog(turn : DialogueTurn) {
+        this._log.push(turn);
+    }
+
+    upvoteLast() {
```
gcampax, line 578: One method, `voteLast` with a vote parameter? Also, I'm not a fan of using emojis for this.

lib/dataset-tools/parsers.ts
```diff
@@ -206,14 +210,19 @@ class DialogueSerializer extends Stream.Transform {
...
+                    this._pushMany(this._prefixLines(turn.context, 'C: '));
+                this._pushMany(this._prefixLines(turn.agent, 'A: '));
                 if (this._annotations)
-                    this._pushMany(this._prefixLines(turn.agent_target!, 'AT: '));
+                    this._pushMany(this._prefixLines(turn.agent_target, 'AT: '));
                 if (this._annotations && turn.intermediate_context)
                     this._pushMany(this._prefixLines(turn.intermediate_context, 'C: '));
+
+                if (turn.rate)
+                    this._pushMany(this._prefixLines(turn.rate, '#! '));
```
gcampax, line 221: If we're introducing `#!` to mean "comments that are actually structured data", maybe we should have some sort format inside them, maybe a dictionary or something? So you'd have:
```
#! rate=good timestamp=2021-02-05T... comment=...
```
(it can be even json, although it would be a bit less readable)
(I'm including "comment" in there because I don't quite like conflating the user's comments on the agent reply with any free-form comments of whoever made the file)

lib/dialogue-agent/conversation.ts
```diff
@@ -215,6 +224,18 @@ export default class Conversation extends events.EventEmitter {
         return this._history;
     }
 
+    get inTestMode() : boolean {
```
gcampax, line 227: I mentioned this on Slack, I don't like "test mode", because it's confusing with running inside unit tests or with a simulator. I would rather "recording mode" or similar.

lib/dialogue-agent/conversation.ts
```diff
@@ -342,13 +363,13 @@ export default class Conversation extends events.EventEmitter {
...
 
-    private async _errorWrap(fn : () => Promise<void>, platformData : PlatformData) : Promise<void> {
+    private async _errorWrap(fn : () => Promise<void>, command : string|null, platformData : PlatformData) : Promise<void> {
         try {
             try {
                 await fn();
             } catch(e) {
                 if (this._isUnsupportedError(e))
-                    await this._doHandleCommand(new UserInput.Unsupported(platformData), null, [], true);
+                    await this._doHandleCommand(new UserInput.Unsupported(command!, platformData), null, [], true);
```
gcampax, line 372: This null assertion doesn't look right? It seems from the type change you just made that `command` can indeed be null at the beginning of the function, and there is nothing that prevents it being null here.

lib/dialogue-agent/user-input.ts
```diff
@@ -56,26 +57,29 @@ function parseSpecial(intent : Ast.SpecialControlIntent, context : { platformDat
...
-                return new UserInput.Answer(thingtalk.intent.value, context.platformData);
+                return new UserInput.Answer(thingtalk.intent.value, context.command, context.platformData);
             else if (thingtalk.intent instanceof Ast.ChoiceControlIntent)
-                return new UserInput.MultipleChoiceAnswer(thingtalk.intent.value, context.platformData);
+                return new UserInput.MultipleChoiceAnswer(thingtalk.intent.value, context.command, context.platformData);
             else
                 throw new TypeError(`Unrecognized bookkeeping intent`);
         } else if (thingtalk instanceof Ast.Program) {
-            return new UserInput.Program(thingtalk, context.platformData);
+            return new UserInput.Program(thingtalk, context.command!, context.platformData);
```
gcampax, line 80: Yeah, all these null assertions are iffy. Either things can be null, or they can't.

gcampax (changes requested):
There is nothing glaring, but I think there is room for one round of improvement, especially in the file format. And we do need to split the log into multiple dialogues to match the state tracking in ThingTalk.

lib/dialogue-agent/user-input.ts
rayslxu: > Yeah, all these null assertions are iffy. Either things can be null, or they can't.

Different subclasses of `UserInput` have different types. E.g., `Answer` can be generated by UI pickers, so it could be null; while Program can only be generated by either neural network or with `\t` (at least currently) - so there's always an "utterance". Do you think we should just relax it and allow all subclasses to have null utterance?

gcampax: I think so. It's very hard to prove that the code is correct otherwise, and if the code is not correct, you're making the type system unsound which is even worse than no type system at all.

lib/dialogue-agent/conversation.ts
rayslxu: This was actually just used for testing... I'm actually not sure how the download from almond-cloud would be implemented. Any existing code I can refer to?

lib/dialogue-agent/conversation.ts
rayslxu: Are you saying I should have a list of dialogueTurn lists, and save them into separate files? I don't quite follow: when the dialogue resets, `this._dialogueState` is reset, and it is used to compute the new state. what else do I need to do?

gcampax: You should flush the current list of DialogueTurn lists to the DialogueParser when the state resets, so it closes the current dialogue in the file and it starts a new dialogue.
That way the state that you write in the file matches the state that you have in memory.

lib/dialogue-agent/conversation.ts
gcampax: Download from almond-server is easy: return a filename, and just stream the file content in the download API with `fs.createReadStream(filename).pipe(res)`. From almond-cloud is tricky because the frontend processes might not have access to the same filesystem where the backend process resides. You would need some sort of streaming protocol.

lib/dataset-tools/parsers.ts
gcampax: Until we start tracking timestamps or other metadata about the dialogue...

rayslxu:
> There is nothing glaring, but I think there is room for one round of improvement, especially in the file format. And we do need to split the log into multiple dialogues to match the state tracking in ThingTalk.

I think I have addressed all the comments. 

As for the structured comments. I still only have `vote` and `comment`, and the syntax is as follows: 
```
#! vote: up
#! comment: test comment for dialogue turns
#!          additional
#!          lines
```

lib/dialogue-agent/conversation.ts
```diff
@@ -571,40 +597,61 @@ export default class Conversation extends events.EventEmitter {
...
-        const last = this._log[this._log.length - 1];
+        const last = this.lastTurn();
         last.comment = comment;
     }
 
     async saveLog() {
-        const dir = this._engine.platform.getWritableDir();
+        const dir = path.join(this._engine.platform.getWritableDir(), 'logs');
+        if (!fs.existsSync(dir))
+            fs.mkdirSync(dir);
```
gcampax, line 637: Technically, this is bad practice: you should never check that a file exists before doing something. You should try to make the directory, and ignore the error if it's EEXIST. This avoids a race condition (also known as Time-Of-Check-To-Time-Of-Use vulnerability). Here, it's harmeless, but there is no reason to write bad code.

gcampax (approved):
Let's get this in when tests pass.

PR #453, opened 2021-02-09
Refactor how entities are generated in agent sentences

### LINKED ISSUE #447: Choose better units of measurement in agent responses
This is one example from MultiWOZ dataset (dialogue `# SNG02346.json`):

```
...
U: can i get the price and time travel please
A: Tr3447 is a train with duration 2820000 ms and cost gbp 9.80. Would you like to book it?
```
The value in response should be in hours, not ms.

gcampax:
The goal of this refactor is two fold:
- We need to generate delexicalized agent utterances at inference
  time to pass to the NLG
- We want the actual text of the entity to be generated by the I18n
  code during postprocessing, so consistent postprocessing is
  applied to both synthetic and NLG sentences, and we can use
  locale-specific code.

To apply this refactoring, the system of inference time constants
is modified to make use of a SequentialEntityAllocator, which allocates
entity tokens (NUMBER_0, DATE_1, etc.) for each value that appears
in the context. This entity allocator is also passed to the Describe
module when that module extracts a value from the program.

At training time, we have only one SequentialEntityAllocator,
which is primed with the fake entities used during generation,
plus any entity that is generated by the simulator and observed
by Describe. At inference time, we reset the entity allocator
immediately prior to processing a new context.

The entities thus allocated are later passed to the I18n module for
postprocessing. This module gained code to automatically select
the most appropriate unit for a given size.

Fixes #447

### DIFF
--- a/languages/thingtalk/load-thingpedia.ts
+++ b/languages/thingtalk/load-thingpedia.ts
@@ -157,7 +157,8 @@ export default class ThingpediaLoader {
         this._ttUtils = ttUtils;
         this._grammar = grammar;
         this._langPack = langPack;
-        this._describer = new ttUtils.Describer(langPack.locale, options.timezone, options.forSide);
+        this._describer = new ttUtils.Describer(langPack.locale,
+            options.timezone, options.entityAllocator, options.forSide);
 
         this._tpClient = options.thingpediaClient;
         if (!options.schemaRetriever) {
--- a/lib/dialogue-agent/abstract_dialogue_agent.ts
+++ b/lib/dialogue-agent/abstract_dialogue_agent.ts
@@ -22,6 +22,7 @@
 import assert from 'assert';
 import { Ast, SchemaRetriever } from 'thingtalk';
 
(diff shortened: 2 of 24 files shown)


s-jse (approved):
Does this affect the synthesis of training data as well?

gcampax:
> Does this affect the synthesis of training data as well?

Only if there is a bug. That is, synthesis was delexicalized before and should remain delexicalized afterwards.

s-jse:
> 
> 
> > Does this affect the synthesis of training data as well?
> 
> Only if there is a bug. That is, synthesis was delexicalized before and should remain delexicalized afterwards.

In the dataset we use to train semantic parsers for agent utterances, agent utterances are not delexicalized. Will this PR affect them?

gcampax:
> In the dataset we use to train semantic parsers for agent utterances, agent utterances are not delexicalized. Will this PR affect them?

Those sentences are delexicalized wrt numbers dates measurements etc. That's the focus of this PR. Augmentation for quoted strings is a separate process.

PR #455, opened 2021-02-09
Wip/mehrad/next

Mehrad0711:


lib/i18n/tokenizer/base.ts
```diff
@@ -524,7 +530,7 @@ export default class BaseTokenizer {
         });
     }
 
-    tokenize(text : string) : TokenizerResult {
+    tokenize(text : string, preserved ?: RegExp) : TokenizerResult {
```
gcampax, line 533: What's this about?

lib/utils/entity-utils.ts
```diff
@@ -26,7 +26,7 @@ export type EntityMap = Syntax.EntityMap;
 
 const MAX_SMALL_INTEGER = 12;
 
-const ENTITIES : { [key : string] : (idx : number) => AnyEntity } = {
+export const ENTITIES : { [key : string] : (idx : number) => AnyEntity } = {
```
gcampax, line 29: Why do you need to export this? Use `makeDummyEntities`, don't use this directly.

tool/typecheck.ts
```diff
@@ -198,6 +207,13 @@ class TypecheckStream extends Stream.Transform {
                 schemaRetriever: this._schemas,
             }, true);
 
+            const preserved_tokens : string[] = Object.keys(EntityUtils.ENTITIES);
+            preserved_tokens.push('GENERIC_ENTITY');
```
gcampax, line 211: You're missing MEASURE here.

tool/typecheck.ts
```diff
@@ -198,6 +207,13 @@ class TypecheckStream extends Stream.Transform {
...
+            const preserved_regex = new RegExp(preserved_tokens.join("|"));
+
+            if (!this._tokenized)
+                this._current.preprocessed = this._tokenizer.tokenize(this._current.preprocessed, preserved_regex).rawTokens.join(' ');
```
gcampax, line 215: What's going on here? The datasets that this tool processes are already tokenized.

gcampax (changes requested):
Some questionable things here.

lib/i18n/tokenizer/base.ts
```diff
@@ -199,7 +202,10 @@ export default class BaseTokenizer {
         // numbers are considered part of a word if preceded by a letter
         // if a word ends with hyphen followed by a number, the hyphen is considered part of the word: e.g., top-50 -> top- 50
         // if a word ends with hyphen, the hyphen is not considered part of the word: e.g., Twitter- -> Twitter -
-        this._addDefinition('WORD', /(?:-?{LETTER}[0-9۰-۹]*)+(?:-?(?=[0-9۰-۹]))?/);
+        // updates:
+        // ... except for -lrb- and -rrb-
+        // person first names or abbreviations (such as st.) may end with a dot; do not split them
```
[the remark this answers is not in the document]
Mehrad0711: These commits are mainly the result of my experience with typechecking SPL and autoQA datasets to port them to thingtalk2. 
If the sentence contains abbreviations like "st.", "dr.", etc. they won't typecheck.
It might be better to not split them unless they are period at the end of a sentence?

lib/i18n/tokenizer/base.ts
Mehrad0711: If we want to make sure a token doesn't get tokenized - similar to special_tokens in huggingface tokenizers.
(I found it useful to keep entities such as "PHONE_NUMBER" intact )

tool/typecheck.ts
Mehrad0711: Don't you want to have the option though?
I found it useful to retokenize spl datasets where tokenization was done with previous code. More sentences typechecked as a result.

lib/i18n/tokenizer/base.ts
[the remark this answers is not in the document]
gcampax: Yes the issue of course is that you have no idea if something is a period at the end of the sentence or not, if a single command is made of two sentences!

tool/typecheck.ts
gcampax: That's not really a use case we want to support, no. Tokenizing an existing tokenized dataset is not guaranteed to be idempotent, or even do anything sensible.

lib/i18n/tokenizer/base.ts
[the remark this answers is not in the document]
Mehrad0711: Actually, I just noticed we have _initAbbrv() and those tokens should go there...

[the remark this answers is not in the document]
gcampax: Yes if you want to extend the list of abbreviations for English (without making every word followed by . into an abbreviation). But I tried to keep the list smaller than the old CoreNLP-based tokenizer because the more we do the more we risk making mistakes.

gcampax (approved):
Okay

PR #459, opened 2021-02-10
Use a postprocessing instead of training to add [1] clause to playing music

gcampax:
When we translate "play X by Y", we add [1] to the overall query, so that only the top matching song is played. We currently do it during synthesis, but I found that the model is confused by which examples need [1] and which don't, despite being a simple rule that can be learned. This PR attempts to achieve that as a postprocessing on the program before execution instead, which should remove some confusion in the model and improve accuracy.

Marking this as draft as I test this on Kubeflow.


gcampax:
This is yielding almost 7% on Spotify (compare gcampax/90 and gcampax/93 in Kubeflow), so yeah I'm going to take that.

PR #461, opened 2021-02-10
Disable support for multidomain contexts

gcampax:
This is a large hammer to solve some of the batch size issues we're
seeing in the Thingpedia model, as we generate contexts that are
too long. With this change, the context will contain at most
one statement, and therefore will be quite short, at the expense
of not supporting any multi-domain parameter passing.

This commit is incompatible with MultiWOZ and will need to be
reverted or made into an option.

@s-jse comments? I can't see any other way to avoid continuous batch size problems.

### DIFF
--- a/languages/thingtalk/en/dlg/search-questions.genie
+++ b/languages/thingtalk/en/dlg/search-questions.genie
@@ -226,13 +226,15 @@ specific_dontcare_phrase : C.FilterSlot = {
 }
 
 coref_constant : Ast.Value = {
-    ( ctx:ctx_multidomain ('the' | 'the same') base:base_table '\'s' param:out_param_Any with { functionName = base.functionName } '.'
-    | ctx:ctx_multidomain generic_preamble_for_answer ('the' | 'the same') base:base_table '\'s' param:out_param_Any with { functionName = base.functionName } '.'
-    | ctx:ctx_multidomain 'the same' param:out_param_Any 'as' 'the' base:base_table with { functionName = param.functionName } '.'
-    | ctx:ctx_multidomain generic_preamble_for_answer 'the same' param:out_param_Any 'as' 'the' base:base_table with { functionName = param.functionName } '.'
-    | ctx:ctx_multidomain ('the same as the' | 'the') param:out_param_Any 'of' 'the' base:base_table with { functionName = param.functionName } '.'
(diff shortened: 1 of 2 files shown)


s-jse:
Batch size problems? What batch size problems?

gcampax:
We keep hitting CUDA OOM errors, and we keep lowering the batch size. But the batch size is ~11 now, which is too low and unsustainable because we don't even see the whole dataset in one training.

s-jse:
> 
> 
> We keep hitting CUDA OOM errors, and we keep lowering the batch size. But the batch size is ~11 now, which is too low and unsustainable because we don't even see the whole dataset in one training.

And what would be the batch size after this change?

gcampax:
Batch size should be the same I think, but the number of examples in each batch should be larger because each example will have a shorter context on average. Although tbqh I am not sure why we hit CUDA OOM given that we specify the batch size in tokens and not in examples.

s-jse:
Batch size is specified by the number of tokens **in the input**, but the memory consumption on GPU is roughly a function of maximum of input and output lengths. The reason we use input length alone for specifying the batch size is that 1) In `TransformerLSTM` models, memory is dominated by the encoder, so input length is the only thing that matters, and 2) length of the output is not known in some cases since we don't have access to the gold output (or even if we did, generated output can have a different length than the gold output).
So two batches that both fit into the specified size limit can have different output lengths, and therefore one of them may result in OOM.

gcampax:
This is not very efficient. Can we change the specification to account for input and output length? The only case where we don't know the output length is when running in the server, but batching for the server is done client side so it doesn't matter.

s-jse:
> 
> 
> This is not very efficient. Can we change the specification to account for input and output length? The only case where we don't know the output length is when running in the server, but batching for the server is done client side so it doesn't matter.

I opened an issue for this in genienlp: https://github.com/stanford-oval/genienlp/issues/91.
Anyways, for this PR I would say let's measure the difference in the maximum number of examples we can fit into a batch after this change to see if it is significant. You can merge this, I just want to know the number.

s-jse (approved):

gcampax:
I started job 95, I will let you know.

gcampax:
It's a bit difficult to notice the number, but I think it oscillates between 10 and 13, with the occasional 14.

gcampax:
It doesn't seem to affect accuracy, or affect batch sizes too much, so I'm not sure I'll merge this after all.

gcampax:
We're not going to merge this after all.

PR #467, opened 2021-02-12
update kf-server json parse

jgd5:
remove extra JSON parse since genienlp has been updated.


gcampax:
I had a similar fix in #421 that I just merged, I think this is superseded now.

jgd5:
L

> I had a similar fix in #421 that I just merged, I think this is superseded now.

Looks good!

PR #469, opened 2021-02-12
Fix logging: update the log whenever agent/user says something

rayslxu:
Always update the actual log directly and use an extra field to track if the current turn is done.


gcampax (approved):
Ok

PR #475, opened 2021-02-15
builtin: fix random number if low > 6

### LINKED ISSUE #474: Random number generator has an implementation bug
> User: give me a random number greater than 10
> Almond: 8.
> User: give me a random number less than 10
> Almond: 9.

Logs show that both user utterances were parsed to this:
`Analyzed message into $dialogue @org.thingpedia.dialogue.transaction . execute ; @org.thingpedia.builtin.thingengine.builtin . get_random_between ( low = 10 ) ;`

gcampax:
If low > high (which defaults to 6), the existing math is incorrect.
Instead, if low > high, raise the high to low + 6, keeping the default
range in case the user only specifies high.

Fixes #474

### DIFF
--- a/lib/engine/devices/builtins/thingengine.builtin.ts
+++ b/lib/engine/devices/builtins/thingengine.builtin.ts
@@ -58,10 +58,16 @@ export default class MiscellaneousDevice extends Tp.BaseDevice {
         // FIXME convert to the right timezone...
         return [{ time: new Tp.Value.Time(now.getHours(), now.getMinutes(), now.getSeconds()) }];
     }
-    get_get_random_between({ low, high } : { low : number, high : number }) {
-        low = (low === null || low === undefined) ?  1 : low;
-        high = (high === null || high === undefined) ?  6 : high;
-        return [{ random: Math.round(low + (Math.random() * (high - low))) }];
+    get_get_random_between({ low, high } : { low : number|null|undefined, high : number|null|undefined }) {
+        if ((low === null || low === undefined) && (high === null || high === undefined)) {
+            low = 1;
+            high = 6;
+        } else if (low === null || low === undefined) {
+            low = high! < 0 ? Math.min(2 * high!, high! - 5) : 1;
+        } else if (high === null || high === undefined) {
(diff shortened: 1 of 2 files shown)


rayslxu (approved):
It's a little weird to get only numbers between 100 and 106 when user asking a random number greater than 100. But let's at least don't generate the wrong number first.

gcampax:
Do you have a different suggestion for how to choose an upper bound if only a lower bound is specified?

Also, one quirk of the old code is that the math works out correctly if you swap high and low (ie, you always get a number between the smaller and the larger of the two numbers). This quirk papers over bad parses.
With this change, the quirk is no more.

s-jse (changes requested):
This does not support negative numbers: the default low should not be 1, if it is not specified but high is specified, it should be determined using the default range.

And a slightly better default range is 2*abs(low). So "a random number above 100" will be between (100, 200) and "a random number above -10" will be (-10, 10).

gcampax:
@s-jse take a look at the revised implementation, see if this makes more sense to you.

lib/engine/devices/builtins/thingengine.builtin.ts
```diff
@@ -58,10 +58,16 @@ export default class MiscellaneousDevice extends Tp.BaseDevice {
...
     }
-    get_get_random_between({ low, high } : { low : number, high : number }) {
-        low = (low === null || low === undefined) ?  1 : low;
-        high = (high === null || high === undefined) ?  6 : high;
-        return [{ random: Math.round(low + (Math.random() * (high - low))) }];
+    get_get_random_between({ low, high } : { low : number|null|undefined, high : number|null|undefined }) {
+        if ((low === null || low === undefined) && (high === null || high === undefined)) {
+            low = 1;
+            high = 6;
+        } else if (low === null || low === undefined) {
```
s-jse, line 65: Maybe something like this?
```
else if (low === null || low === undefined) {
            low = high! < 0 ? min(2 * high!, high!-5) : 1;
        } else if (high === null || high === undefined) {
            high = low < 0 ? 0 : max(2 * low, low+5);
        }
```

s-jse (approved):

Issue #483, opened 2021-02-19
Speech output should not read out the content of cards


gcampax:
Cards (RDL, pictures) are for graphical consumption only, and will always be preceded with a fully self-contained text message. Hence, the speech output should not read them.

PR #491, opened 2021-02-22
Assorted bug fixes

gcampax:
See individual commits

### DIFF
--- a/languages/thingtalk/en/basic.genie
+++ b/languages/thingtalk/en/basic.genie
@@ -55,6 +55,20 @@ bookkeeping : Ast.ControlCommand = {
     | 'close'
     ) => special('stop');
 
+    ( 'yes'
+    | 'yes please'
+    | 'yes ,  please'
+    | 'yeah'
+    | 'ok'
+    | 'okay'
+    ) => special('yes');
+
+    ( 'no'
+    | 'no thanks'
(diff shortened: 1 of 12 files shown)


rayslxu (approved):
Looks good. 
I'm not sure about #483 though - pretty much all the news apis return RDL with title and we probably want to read that. the fix is fine for the skills we have right now, but in the future, we need something smarter.

rayslxu:
For #483

> Looks good.
> I'm not sure about #483 though - pretty much all the news apis return RDL with title and we probably want to read that. the fix is fine for the skills we have right now, but in the future, we need something smarter.

gcampax:
You don't want to read the news title straight up, you want to construct a sentence that contains the news title, using the `#[result]` phrase, and read that. So the RDL is always redundant. At least, that's what the SmartNews skill does.

rayslxu:
> You don't want to read the news title straight up, you want to construct a sentence that contains the news title, using the `#[result]` phrase, and read that. So the RDL is always redundant. At least, that's what the SmartNews skill does.

So in text mode, there will be a phrase as well as the RDL link with redundant information?

gcampax:
Yes. I believe that's the design on Google Assistant as well, judging from their documentation around card outputs.

Issue #484, opened 2021-02-22
Big numbers in string values are parsed as special token NUMBER_x


rayslxu:
3 examples in schema2qa eval set are removed due to this issue. See https://github.com/stanford-oval/genie-toolkit/pull/472/files

s-jse:
Just to clarify, this constitutes a bug. Currently, users cannot ask about books, movies, etc. that contain large numbers in their title.

PR #492, opened 2021-02-23
No more number preprocessing

gcampax:
This is a draft of the nuclear option to solve #484 (and several related bugs marked as "bug-tokenize" in various eval sets).

With this PR, the tokenizer recognizes all numbers as before, but replaces them with a normalized digit form instead of replacing with NUMBER tokens. The normalized form is slightly locale-specific (e.g. it uses , instead of . as digit separator in Italian) so that it is possible to preprocess the normalized form without loss of information.

This PR only affects NUMBER tokens (numbers and measurements). The tokenizer still replaces:
- Times and dates
- Currencies
- Phone numbers
- Email addresses
- Usernames (words that start with @) and hashtags (words that start with #)
- URLs and Unix/Windows path names

We previously discussed dropping CURRENCY tokens as well, so that shoudn't be controversial. 
Times and dates are trickier because the normalized form would be a lot more complex and very locale specific, and the neural model likely would need some complex logic to copy the normalized date bits into a `new Date` form in ThingTalk.
As for the rest, it is less obvious to me that they benefit from the neural model seeing them, and results in Nancy's paper suggest that masking URLs is very useful.

Marking this PR as draft because I need to test it with Kubeflow, but early feedback welcome. Please do test this with Schema.org as well.


gcampax:
I tested this on Kubeflow, we get about a 1% reduction in exact match. Interestingly, it seems to be mostly from Spotify and Yelp commands that do not use numbers. I think the model struggles to learn too many things at once, and when it has to focus on copying numbers it starts losing track of some other part of the sentence. Regardless, a 1% loss is probably offset by the sentences that we lose due to preprocessing, so this is acceptable.

gcampax:
Mh on a closer look at Travis, it seems that we generate some NaNs in the augmentation, which is not good. Not sure how reliable are the Kubeflow results then.

rayslxu (approved):

s-jse:
> 
> Mh on a closer look at Travis, it seems that we generate some NaNs in the augmentation, which is not good. Not sure how reliable are the Kubeflow results then.

So do we have the correct accuracy number after this is fixed?

gcampax:
The issue only affected constants of temperature type, which we have in the tests but don't have in the Thingpedia model. So the numbers ought to be correct already.

PR #502, opened 2021-02-26
Another round of bug fixes

gcampax:
Which, among other, fix synthesis at full size on the Thingpedia model


rayslxu (approved):
looks good

PR #505, opened 2021-02-27
Bug fix: actually disable recording when it is disabled

### LINKED ISSUE #503: Recording is not actually stopped after calling `endRording`.

rayslxu:
Only update logs when in recording mode;
Also mark dialogue as finished immediately after recording is disabled,
so the next time recording gets enabled, we do not append to the last
unfinished dialogue.

Fix #503


gcampax (approved):
Makes sense

PR #510, opened 2021-03-01
DeviceDatabase: implement automatic device updates

gcampax:
Run a timer to check for new versions of the skills every few
hours. That way, any bug fixes that are deployed to Thingpedia
become available to testers without me clearing the caches and
restarting the engines.

### DIFF
--- a/lib/engine/devices/database.js
+++ b/lib/engine/devices/database.js
@@ -25,6 +25,11 @@ import * as Tp from 'thingpedia';
 const ObjectSet = Tp.ObjectSet;
 import SyncDatabase from '../db/syncdb';
 
+// check for updates of all devices every 3 hours
+// this is to catch bug fixes quickly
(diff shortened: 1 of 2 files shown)


lib/engine/devices/database.js
```diff
@@ -25,6 +25,11 @@ import * as Tp from 'thingpedia';
 const ObjectSet = Tp.ObjectSet;
 import SyncDatabase from '../db/syncdb';
 
+// check for updates of all devices every 3 hours
+// this is to catch bug fixes quickly
+// we'll lower this to 24 hours after the alpha release
+const UPDATE_FREQUENCY = 60 * 1000; //3 * 3600 * 1000;
```
rayslxu, line 31: isn't this 1 minute?

gcampax: Ugh yes I set this to a minute for testing, I forgot to revert back to 3 hours

PR #516, opened 2021-03-03
Update ThingTalk to git

gcampax:
To test the latest version before I make another release, because
the release will contain existential subqueries, which can potentially
break a lot of things.


gcampax:
Ok, I generated a full size dataset and it seems to be approximately the same, so I'm going to assume things will work correctly with the new ThingTalk now.

PR #517, opened 2021-03-04
Add a dialogue parser mode which tolerates errors

gcampax:
Dialogues generated by the recording mode are not fully
well-formed dialogue training samples because of commands that
go outside the state machine (such as "$stop;", "$failed;") and
because of replies that terminate the dialogue (so the last
turn is always incomplete)

### DIFF
--- a/lib/dataset-tools/parsers.ts
+++ b/lib/dataset-tools/parsers.ts
@@ -255,8 +255,13 @@ class DialogueParser extends Stream.Transform {
     private _i : number;
     private _id : string|undefined;
     private _keySequence : Array<keyof DialogueTurn>;
+    private _ignoreErrors : boolean;
 
-    constructor({ withAnnotations = true, invertTurns = false } = {}) {
+    constructor({
+        withAnnotations = true,
+        invertTurns = false,
+        ignoreErrors = false
+    } = {}) {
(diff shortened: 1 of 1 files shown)


lib/dataset-tools/parsers.ts
```diff
@@ -405,21 +413,36 @@ class DialogueParser extends Stream.Transform {
                 if (key === 'context' && this._keySequence[expect] === 'user')
                     key = 'intermediate_context';
 
+                if (this._ignoreErrors) {
+                    if (key === 'agent' && this._keySequence[expect] === 'context') {
+                        currentTurn.context = currentTurn.user_target;
+                        expect = (expect + 1) % this._keySequence.length;
```
rayslxu, line 419: what does this do?

gcampax: Basically, sometimes in the recording you have a sequence:
```
U: foo
UT: $failed;
A: Sorry I don't know how to do that yet
...
```
(or similar)

This is malformed because between UT and A you should have C
Those lines "adjust" the context to be:
```
U: foo
UT: $failed;
C: $failed;
A: Sorry I don't know how to do that yet
...
```
And then we can keep going.

rayslxu (approved):

PR #530, opened 2021-03-08
Fix prefix operator subquery in schemaorg starter code

rayslxu:


gcampax (approved):
Yeah seems reasonable

PR #542, opened 2021-03-13
Synthesis refactoring, part N

gcampax:
Implementing the design in https://wiki.almond.stanford.edu/en/genie/design-notes/new-template-language and finally having the infrastructure for good rule-based NLG in all languages.


gcampax:
I'm happy with the result so far, because it solves the most pressing grammar issues. We can continue migrating the templates to the new syntax as needed.

rayslxu (approved):
Love it. already make quite a lot of code much cleaner.

PR #543, opened 2021-03-14
Basic subquery templates

rayslxu:
This PR converts existing templates for joins and get predicate into subquery templates. A separate PR will be submitted to improve the templates for subqueries. 

Some SPL changes broke the schema.org pipeline. @Mehrad0711, let me know if you still want to have them in master.


languages/thingtalk/ast_manip.ts
```diff
@@ -1467,54 +1401,50 @@ function addGetPredicateJoin(table : Ast.Expression,
...
-        get_predicate_table.expression.invocation.selector,
-        get_predicate_table.expression.invocation.channel,
-        get_predicate_table.expression.invocation.in_params,
-        new Ast.BooleanExpression.And(null, [get_predicate_table.filter, newAtom]),
-        get_predicate_table.expression.invocation.schema
+    // add id projection to subquery
+    if (subquery instanceof Ast.ProjectionExpression)
+        subquery.args = ['id'];
+    else
+        subquery = new Ast.ProjectionExpression(null, subquery, ['id'], [], [], subquery.schema);
```
gcampax, line 1408: Wait is this the right form? `[id]` is a minimal projection, wouldn't it be optimized out if projected explicitly?

languages/thingtalk/ast_manip.ts
```diff
@@ -1467,54 +1401,50 @@ function addGetPredicateJoin(table : Ast.Expression,
...
+
+    const comparisonSubquery = new Ast.BooleanExpression.ComparisonSubquery(
+        null,
+        new Ast.Value.VarRef(lhsArg.name),
+        lhsArg.type.isArray ? 'contains' : '==',
```
gcampax, line 1413: Should this be `contains~` and `=~` if the type is string?

languages/thingtalk/ast_manip.ts
```diff
@@ -1467,54 +1401,50 @@ function addGetPredicateJoin(table : Ast.Expression,
...
-        newAtom = new Ast.BooleanExpression.Not(null, newAtom);
-    const get_predicate = new Ast.BooleanExpression.External(null,
-        get_predicate_table.expression.invocation.selector,
-        get_predicate_table.expression.invocation.channel,
-        get_predicate_table.expression.invocation.in_params,
-        new Ast.BooleanExpression.And(null, [get_predicate_table.filter, newAtom]),
-        get_predicate_table.expression.invocation.schema
+    // add id projection to subquery
+    if (subquery instanceof Ast.ProjectionExpression)
+        subquery.args = ['id'];
```
gcampax, line 1406: Do you need to clone subquery before you modify it like this?

languages/thingtalk/ast_manip.ts
```diff
@@ -1552,10 +1482,9 @@ function makeComputeFilterExpression(table : Ast.Expression,
 function makeWithinGeoDistanceExpression(table : Ast.Expression, location : Ast.Value, filterValue : Ast.Value) : Ast.Expression|null {
     if (!table.schema!.out.geo || !table.schema!.out.geo.isLocation)
         return null;
-    const filterType = filterValue.getType();
-    if (!(filterType instanceof Type.Measure))
+    if (!(filterValue instanceof Ast.Value.Measure))
         return null;
-    const unit = filterType.unit;
+    const unit = filterValue.unit;
```
gcampax, line 1487: Why this change? I don't think it's correct if you have a sum of two measurements.

tool/autoqa/schemaorg/manual-annotations.js
```diff
@@ -592,21 +594,16 @@ const STRING_FILE_OVERRIDES = {
     'org.schema.Restaurant:Restaurant_name': 'org.openstreetmap:restaurant',
     'org.schema.Person:Person_name': 'tt:person_full_name',
     'org.schema.Person:Person_address_addressLocality': 'tt:location',
-    'org.schema.Hotel:Hotel_geo': 'org.schema.Hotel:Hotel_address_addressLocality',
```
Mehrad0711, line 595: I think this change goes back to the problem of not being able to specify distinct parameter files for each language. Do you need to do it this way? If not, I prefer to retain the old approach.

rayslxu: You broke schema2qa experiment with this. We don't do filters on street/city in schema2qa. 
If you want to use this script to generate a different schema, you should make it an option for SPL experiments.

languages/thingtalk/ast_manip.ts
rayslxu: I notice you changed this in a previous commit. But the `filterType` unit is always 'm' no matter what's the value's type. We end up generate no within filters.

gcampax: Why do we not generate filters if the unit is m? m is a valid distance unit...

languages/thingtalk/ast_manip.ts
rayslxu: Seem not the case. where is the code to check minimal projection? didn't find it in optimize.ts

languages/thingtalk/ast_manip.ts
rayslxu: yes, I will fix this

languages/thingtalk/ast_manip.ts
rayslxu: when I debug, I find that no matter what the original uni is, the unit from `filterType` is always `m`, and we don't generate numbers over 100

languages/thingtalk/ast_manip.ts
rayslxu: what issue will it cause without clone?

gcampax: The same object that you pass in, used in a different template, will be modified and cause havoc.

languages/thingtalk/ast_manip.ts
gcampax: Ok so is the problem with min_number/max_number? Those need special handling of "large" numbers, which we generate by proxy generating numbers between 13 and 23

languages/thingtalk/ast_manip.ts
rayslxu: oh actually this basic version only generate id filter, so this is not an issue for this pr

languages/thingtalk/ast_manip.ts
rayslxu: No, the issue is that in the following, when we check if the unit in the utterance is at least feet, or the distance is at least 100m, we are not using the actual unit. 
do we really need to have "within a + b miles"?

gcampax: We don't, but I feel like if we don't handle that case correctly it will cause hard to debug issues later on.

rayslxu: this code is only for "within" filter; if it's a computation value, it won't break it will just return null, as it should be, because it is not very useful sentence. why will it cause problems in the future...

tool/autoqa/schemaorg/manual-annotations.js
Mehrad0711: Not sure what you mean by "broke", cause the tests were passing when changes were merged and it was reviewed by you and Giovanni.
Change it however you want. I will take a look at SPL experiments later.

rayslxu: We do not do filters on streetAddress or addressLocality (city) in all the annotated data of schema2qa dataset. In fact, they are not even in the schema. Your changes broke that and make them appear in the schema and generate training data filtering on them instead of geo. It won't break any test but will get a very low accuracy on schema2qa eval and test.  

I suppose you modified those eval/test examples in SPL?

languages/thingtalk/ast_manip.ts
gcampax: It might be a normalization that we apply only to synthesized data, and not to all programs. In which case it's fine if this the normalized form you want.

gcampax (approved):
Ok for me

tool/autoqa/schemaorg/manual-annotations.js
Mehrad0711: I do have parameter files for streetAddress and addressLocality for each language, and I was generating new schemas for SPL experiments. Feel free to merge this, and I will test SPL with genie master. If it broke SPL, we can discuss how to change this later.

gcampax: I think the right solution is that all these overrides should be gated by the locale parameter, so you can generate a locale-specific schema file appropriate for SPL. Because using the English manifest might not be correct in other languages for other reasons too.

Mehrad0711: Yeah, having one schema per language is a possible solution.

PR #546, opened 2021-03-17
Make primitive templates out of action canonical forms

gcampax:
Similar to how we do it for queries.

This should solve some bad accuracy in IoT actions that have no parameters and no primitive templates.


rayslxu (approved):
yeah

PR #552, opened 2021-03-25
Changes to support SmartNews

gcampax:
(This PR is on top of #551 , so there are duplicated commits. New commits start at "Introduce an AudioController etc.")

This PR includes all the changes needed to make SmartNews work in the most basic form.
The design of SmartNews, at first, will be that the user says "tell me the news", and the skill start reading headlines, one by one, forever until the user says stop. The SN API will include, for each news item, the URL of a pre-synthesized read out of the headline (and maybe the summary in the future).
This PR started as just introducing the AudioController object to coordinate between SmartNews playing news and Spotify playing music. It evolved in some deeper changes so we can keep "tell me the news" as a query, not an action. This is necessary to get all the filtering/sorting/search capabilities that we like from ThingTalk.

The current design is that SmartNews `articles` query will be declared as:
```
#_[result={ list=["Here are the news"] }]
#_[formatted=[
  { type="rdl", webCallback="${url}", displayTitle= "${title}" },
  { type="audio", url="${audio_url}" }
]
```
On a speech platform, it will read out "Here are the news", and then spawn a player with all the audio URLs returned by the API and filtered by ThingTalk.
On a graphical platform, it will show "Here are the news", and then show the RDLs for all news items.
At any point during playback, the user can say "stop" and the agent will stop the player and go quiet.

Overall, this design works but I'm not convinced it's the right one. So I would take a review and/or comments on the design though.
In particular, still unresolved:
1. cleaning up "stop" (`$stop`) vs "stop speaking" (`set_voice_output(off)`) vs "stop playing" (`@spotify.player_pause`)
2. referring to the news article that the agent is currently speaking with "skip", "pocket this" or "tell me more about this"
3. continuing with more news if the agent runs out and the user did not say stop (continuous playback mode)
4. testing this end-to-end


gcampax:
I tested this with https://github.com/stanford-oval/almond-server/tree/wip/media-player and https://github.com/stanford-oval/thingpedia-common-devices/tree/wip/smartnews-audio. It kind of works. Audio is played back to back with no pause, so I added a sound effect in between (I used "bell" for lack of a better one, but that's not good and won't be the final sound we use).

Still, I'm not convinced this is the right approach, because a lot of this logic seems SmartNews specific. I would want to find a way to delegate the actual playing to SmartNews if possible. Something like the really old `formatEvent` method perhaps?

PR #554, opened 2021-03-30
Fix requoting bug

s-jse:
In the new NN syntax, not all tokens that start with a capital letter are context entities.

### DIFF
--- a/tool/requote.js
+++ b/tool/requote.js
@@ -395,7 +395,7 @@ function requoteSentence(id, context, sentence, program, mode, handle_heuristics
     let contextEntities = new Set;
     if (context) {
         for (let token of context.split(' ')) {
-            if (/^[A-Z]/.test(token))
+            if (/^[A-Z].*_[0-9۰-۹]+/.test(token)) // match strings like QUOTED_STRING_0, NUMBER_0 and GENERIC_ENTITY_uk.ac.cam.multiwoz.Hotel:Hotel_0
                 contextEntities.add(token);
         }
     }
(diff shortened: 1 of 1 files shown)


tool/requote.js
```diff
@@ -396,7 +396,7 @@ function requoteSentence(id, context, sentence, program, mode, requote_numbers,
     let contextEntities = new Set;
     if (context) {
         for (let token of context.split(' ')) {
-            if (/^[A-Z]/.test(token))
+            if (/^[A-Z].*_[0-9|۰-۹]/.test(token)) // match strings like QUOTED_STRING_0, NUMBER_0 and GENERIC_ENTITY_uk.ac.cam.multiwoz.Hotel:Hotel_0
```
gcampax, line 399: The syntax is wrong (it will match the literal character `|` instead of matching either latin or persian digits). But why do you want persian digits here at all?

gcampax (changes requested):

tool/requote.js
gcampax: Also, this is wrong if you have more than 10 entities of a certain type (which can happen in rare cases with dialogues)

s-jse: You are right. I copied from lines 34-36 of the same file, which are wrong too.

gcampax: Yeah but why persian numbers in entities? It doesn't make sense.

s-jse: Yeah, that part is removed in the `wip/mehrad/updates` branch. I removed it here as well.

gcampax:
Unit test failure is unrelated, appears to be https://github.com/stanford-oval/thingtalk/pull/331

s-jse:
> 
> 
> Unit test failure is unrelated, appears to be [stanford-oval/thingtalk#331](https://github.com/stanford-oval/thingtalk/pull/331)

I got worried for a second!
I'm merging this PR then.

gcampax:
I merged an update to ThingTalk that fixes the issue. If you rebase on master tests should pass. You should rebase anyway so you clean up the "Fix" commits.

PR #562, opened 2021-04-07
Merge thingpedia-cli into Genie

gcampax:
The "init-project" and "init-device" commands in thingpedia-cli are always out of date, and do not match our recent best practices. Conversely, we do not have a "starter" package that actually delivers a working custom Genie agent with a Thingpedia skill.
On top of that, having a separate project for thingpedia-cli means that you get versioning issues with the thingtalk and thingpedia libraries, and the "lint-device" command is often out of date and complaining about new annotations.

Let's clean up all of that, so new contributors can start clean.


rayslxu (approved):
yeah, this makes a lot of sense.

PR #565, opened 2021-04-11
Close audio output stream when assistant finishes speaking

ad31c0:
Close audio output stream when assistant finishes speaking.

Emit 'match' event after wakeword confirm.

### DIFF
--- a/lib/speech/speech_synthesizer.ts
+++ b/lib/speech/speech_synthesizer.ts
@@ -48,7 +48,6 @@ export default class SpeechSynthesizer extends events.EventEmitter {
     private _queue : Array<QueueItem|Promise<QueueItem>>;
     private _speaking : boolean;
     private _outputStream : SoundOutputStream|null;
-    private _closeTimeout : NodeJS.Timeout|null;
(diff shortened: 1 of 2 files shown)


lib/speech/speech_synthesizer.ts
```diff
@@ -118,16 +118,19 @@ export default class SpeechSynthesizer extends events.EventEmitter {
...
-            this._outputStream = null;
-            this._closeTimeout = null;
-        }, 60000);
+        this._closeTimeout = null;
+    }
 
+    private _ensureOutputStream(result : { sampleRate : number, numChannels : number }) {
+        if (this._closeTimeout)
+            clearTimeout(this._closeTimeout);
+        this._closeTimeout = setTimeout(this._closeOutputStream, 60000);
```
gcampax, line 133: If we close immediately after we're done speaking we don't need the timeout logic right?

ad31c0: Right, removed all the references

gcampax (approved):
Ok

PR #573, opened 2021-04-14
schema2qa update

rayslxu:
Update evaluation set in starter code with the latest tokenizer.


gcampax (approved):
Nice. Do we recover some of the bug-tokenize dropped eval sentences given the new tokenizer?

rayslxu:
> Nice. Do we recover some of the bug-tokenize dropped eval sentences given the new tokenizer?

I recovered the 3 we dropped lately. I didn't add anything new compared to version 1.0

PR #574, opened 2021-04-15
Fix circular dependency in genie tokenizer

rayslxu:
`./lib/i18n` -> `./lib/i18n/default.ts` -> `./lib/utils/template-strings` -> `./lib/i18n/`

### DIFF
--- a/languages/thingtalk/load-thingpedia.ts
+++ b/languages/thingtalk/load-thingpedia.ts
@@ -818,7 +818,7 @@ export default class ThingpediaLoader {
             }
             return true;
         });
-        parsed.preprocess(this._langPack.locale, names);
+        parsed.preprocess(this._langPack, names);
 
         // template #1: just constants and/or undefined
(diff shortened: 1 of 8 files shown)


gcampax (commented):
This commit causes us to create new Tokenizer instances repeatedly. This is not very efficient because we end up initializing a whole bunch of regular expressions every time.
It's also a potential source of bugs in the future because now there are two sources of truth wrt which languages are supported, one here and one in i18n/index

Is there no other way to resolve the circular import?

gcampax:
E.g. what if we import template-strings from i18n/index before we import default.ts ?

rayslxu:
> E.g. what if we import template-strings from i18n/index before we import default.ts ?

I don't follow. Isn't that still circular? as `./lib/i18n` and './lib/utils/template-strings/` gonna import from each other?

gcampax:
Yes but template-strings doesn't need anything from I18n during the initial module evaluation. So it will receive an empty object (while I18n continues initializing), declare a bunch of classes, and move on happily.

gcampax (approved):
Nice yeah

Issue #368, opened 2020-09-30
Port the device disambiguation dialogue to the state machine


gcampax:
Currently, if the choice of device is ambiguous, we run an [out-of-band dialogue to let the user choose](https://github.com/stanford-oval/genie-toolkit/blob/master/lib/dialogue-agent/execution_dialogue_agent.js#L63). This dialogue is not part of the general state machine, and choosing a device does not go through the neural network. This is suboptimal, because it constrains the way the user can talk about the selected device, and prevents us from having general fallbacks and error handling. Instead, we should fold disambiguation the general state machine.

The challenge with this plan is how to expose the list of available devices to the state machine and to the neural network (ideally, in a privacy-preserving way), both at simulation and at inference time.

gcampax:
Assigning this to myself as a cleanup to be done _after_ #306 and #305

PR #578, opened 2021-04-21
Support for multiple devices during training

### LINKED ISSUE #305: Add templates for choosing among multiple devices
For IoT, done the proper Genie way instead of completely out of band.

### LINKED ISSUE #306: Simulate the existance of multiple devices when generating the dataset
If we don't simulate multiple devices, we never include the device ID in the generated contexts that we use to train the neural network, and that causes a training/runtime mismatch because we do have the device ID at runtime.
This in particular affects Spotify, and probably causes an accuracy drop at second and later turns.

### LINKED ISSUE #540: Commands without a device name should implicitly apply to all devices
"Turn on the lights" should implicitly apply to all the lights, rather than pick one light at random, or ask the user which light to turn on.
I am not sure though if we should apply this to all devices, or this should be an annotation specified on the function/class to control this behavior.

gcampax:
I _think_ this is all that we need to have proper handling of multiple IoT devices.

- We simulate multiple devices at training time
- We propagate the choice of device across multiple turns if necessary
- We disambiguate device names if necessary

Fixes #306 
Fixes #540 

Does not affect #305 or #368 because we still use the disambiguation dialogues outside of the state machine, so the user must give a name when asked and nothing else.
The disambiguation dialogue doesn't work very well in voice because it uses choice buttons, but we'll see how that goes. There might be a follow up PR.

Needs testing in Kubeflow to see how bad we do.

### DIFF
--- a/languages/thingtalk/dialogue_acts/streams.ts
+++ b/languages/thingtalk/dialogue_acts/streams.ts
@@ -45,6 +45,10 @@ export function addStream(ctx : ContextInfo, stream : Ast.Expression) {
     if (currentExpression.first.schema!.functionType === 'stream')
         return null;
 
+    // no self-joins
+    if (C.isSameFunction(currentExpression.schema!, stream.schema!))
+        return null;
+
     return addNewStatement(ctx, 'execute', null, 'accepted',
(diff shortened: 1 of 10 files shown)


gcampax:
After several fixes, this seems to work in Kubeflow.

PR #579, opened 2021-04-21
Fix out-dated evaluate-file script

rayslxu:


tool/evaluate-file.ts
```diff
@@ -61,6 +70,16 @@ export function initArgparse(subparsers : argparse.SubParser) {
         choices: ['thingtalk', 'dlgthingtalk'],
         help: `The programming language to generate`
     });
+    parser.add_argument('--tokenized', {
+        action: 'store_true',
+        help: 'The utterances are tokenized.',
+        default: true
+    });
+    parser.add_argument('--untokenized', {
```
gcampax, line 78: In other files we call this `--no-tokenized`, and we try to keep consistent using `--no-` for negating boolean flags.

tool/evaluate-file.ts
```diff
@@ -40,16 +40,25 @@ export function initArgparse(subparsers : argparse.SubParser) {
...
+    parser.add_argument('--parameter-datasets', {
+        required: true,
+        help: 'Path to parameter dataset manifest.'
+    });
     parser.add_argument('--predictions', {
         required: true,
         type: fs.createReadStream,
-        help: 'Prediction results (in TSV format: id, prediction)'
+        help: 'Prediction results (in TSV format: id, sentence, target, prediction)'
     });
```
gcampax, line 51: I think at some point we wanted this file format to be compatible with `genienlp predict`, which only does `id` and `prediction`. Not sure what happened to that.

tool/evaluate-file.ts
```diff
@@ -84,6 +103,7 @@ export function initArgparse(subparsers : argparse.SubParser) {
     parser.add_argument('--csv', {
         action: 'store_true',
         help: 'Output a single CSV line',
+        default: true
```
gcampax, line 106: You can't default this one to true, because there is no way to turn it off. Other commands also have it default to false.

gcampax (commented):
Ok, but why do you need this script at all?

tool/evaluate-file.ts
rayslxu: not sure, apparently that won't work here...

tool/evaluate-file.ts
rayslxu: > You can't default this one to true, because there is no way to turn it off. Other commands also have it default to false.

oh, yeah, I will revert this. `tokenized` should be fine since there is an negate option, right?

gcampax: Yes. That's basically the only reason to have a negate option...

rayslxu:
> Ok, but why do you need this script at all?

it's for Siva's group. they find it to be easier to dump predictions first than using a server - and we should keep all scripts up-to-date if we keep them in the repo anyway...

gcampax:
No yeah sure we should keep things running. Might be worth adding a test for this script too, it should be easier than testing evaluate-server.

gcampax (approved):

PR #582, opened 2021-04-26
Enable previously disabled templates for dialogues such as subquery

sileix:
- subquery
- count over compound types with filters


gcampax:
It looks like it doesn't die miserably on Travis. Maybe we can enable this?

rayslxu:
> It looks like it doesn't die miserably on Travis. Maybe we can enable this?

yeah, it seems synthesis time is reasonable too. so I suppose I can merge this?

gcampax:
I want to test it on Kubeflow first. I am worried it might create some conflicts with some Spotify commands. I'll merge if it runs without loss of accuracy.

PR #585, opened 2021-04-28
Cherry-pick useful fixes from the Covid Genie work

gcampax:
These are all the changes from the wip/genie-for-covid branch that make sense for normal Almond as well, and should be merged to master to lighten the other PR.


lib/dialogue-agent/dialogue-loop.ts
```diff
@@ -655,10 +665,29 @@ export default class DialogueLoop {
...
     }
 
-    private async _loop(showWelcome : boolean) {
-        // if we want to show the welcome message, we run the policy on the `null` state, which will return the sys_greet intent
-        if (showWelcome)
+    private async _loop(showWelcome : boolean, initialState : string|null) {
+        if (initialState !== null) {
+            if (initialState === 'null') {
+                this._dialogueState = null;
+                await this.setExpected(null);
```
rayslxu, line 684: in what case this is a string 'null'? should we show welcome in this case?

languages/thingtalk/dialogue_acts/empty-search.ts
```diff
@@ -69,6 +63,11 @@ function makeEmptySearchError(ctx : ContextInfo, [base, question] : EmptySearch)
             return null;
         type = arg.type;
         state = makeSimpleState(ctx, 'sys_empty_search_question', [question.name]);
+    } else if (offerMonitor) {
+        const monitor = C.tableToStream(ctx.current!.stmt.lastQuery!);
+        if (!monitor)
+            return null;
+        state = addNewStatement(ctx, 'sys_empty_search', null, 'proposed', monitor);
```
rayslxu, line 70: hmm this is a nice one

lib/dialogue-agent/execution_dialogue_agent.ts
```diff
@@ -382,6 +382,9 @@ export default class ExecutionDialogueAgent extends AbstractDialogueAgent<undefi
     }
 
     private _tryGetStoredVariable(type : Type, variable : string) : Ast.Value|null {
+        if (this._dlg.isAnonymous)
```
rayslxu, line 387: it's probably fine to get variables already stored in anonymous mode?

lib/dialogue-agent/notifications/email.ts
```diff
@@ -67,10 +92,41 @@ export default class EmailNotificationBackend {
...
+            timezone: this._platform.timezone
+        })!;
+
+        let text = data.formatted.map((x) => x.toLocaleString(this._platform.locale)).join('\n');
+        text += `
+----
+You are receiving this email because you subscribed to notifications
+from Genie. To stop, open the following link:
+${unsubscribeURL}
+`;
```
rayslxu, line 108: where does it actually go to? where is the backend to unsubscribe?

lib/dialogue-agent/notifications/twilio.ts
```diff
@@ -40,7 +41,8 @@ export default class TwilioNotificationBackend {
     constructor(engine : Engine, config : TwilioConfig) {
         this._platform = engine.platform;
         this._client = new twilio.Twilio(config.accountSid, config.authToken);
-        this._from = config.from;
+        this._from = config.fromByLocale ? (config.fromByLocale[this._platform.locale] || config.from)
+            : config.from;
```
rayslxu, line 45: just curious, why this feature? someone who has two phone numbers for different languages?

lib/sentence-generator/generator.ts
```diff
@@ -1890,7 +1891,7 @@ function expandRule(charts : ChartTable,
 
     // make an estimate of the number of times we'll need to call the semantic function
     // to get the target pruning size
-    const targetSemanticFunctionCalls = Math.min(targetPruningSize / estimatedPruneFactor, EXPONENTIAL_PRUNE_SIZE);
+    const targetSemanticFunctionCalls = Math.min(targetPruningSize / estimatedPruneFactor, SAMPLING_PRUNE_SIZE);
```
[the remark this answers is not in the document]
gcampax: Basically this occurs where a combination of the semantic function and the constraints cause very few combinations to be valid. Then you end up needing to sample a lot of combinations to fill up the target size.

lib/dialogue-agent/dialogue-loop.ts
gcampax: If initialState is set, it means we're resuming an existing conversation, and we should not show the welcome again. initialState will be the exact string `null` if the agent is not asking a question and the state is null (all commands completed).

lib/dialogue-agent/execution_dialogue_agent.ts
gcampax: It's kind of weird though. Why would you want that?

lib/dialogue-agent/notifications/email.ts
gcampax: In versions of almond-cloud that have the email configuration enabled (aka genie.stanford.edu), there is an endpoint to unsubscribe: https://github.com/stanford-oval/almond-cloud/blob/wip/genie-for-covid/routes/my_api.js#L129

lib/dialogue-agent/notifications/twilio.ts
gcampax: Yes, like Covid Genie, which has +1 866 301 0127 for English, and +1 866 301 0129 for Spanish :stuck_out_tongue:

lib/dialogue-agent/execution_dialogue_agent.ts
rayslxu: for convenience in case that we already know the value? I suppose it's unlikely to switch from unanonymous to anonymous

lib/dialogue-agent/notifications/twilio.ts
rayslxu: oh, this is the agent.. got it

rayslxu (approved):

lib/dialogue-agent/execution_dialogue_agent.ts
gcampax: I mean, you can log in to the "anonymous" user as if it was a real user, and set variables there. I'm just unclear why you would want to do that.

PR #586, opened 2021-04-28
expose language pack to engine

ryachen01:
We need access to the tokenizer for https://github.com/stanford-oval/thingpedia-common-devices/issues/269


gcampax (commented):
This is typescript code, you need to declare the property at the beginning of the class or it won't compile.
You can check that things compile with `npx make`

PR #607, opened 2021-05-02
Another round of bug fixes

gcampax:
Solving the most pressing bugs assigned to me for Almond 2.0


languages/thingtalk/ast_manip.ts
```diff
@@ -1049,14 +1049,14 @@ function builtinSayAction(loader : ThingpediaLoader,
     if (pname instanceof Ast.Value) {
         const param = new Ast.InputParam(null, 'message', pname);
         return new Ast.InvocationExpression(null, new Ast.Invocation(null, selector, 'say', [param], loader.standardSchemas.say),
-            loader.standardSchemas.say.removeArgument('message'));
```
rayslxu, line 1052: why did we have this before?

rayslxu (approved):
looks good

languages/thingtalk/ast_manip.ts
gcampax: In ThingTalk 1.0, with joins and "on" clauses, we used to define the schema for an InvocationAction to be different than the schema of the Invocation: all specified parameters (ie, for actions, all parameters) were removed. It was very confusing and very broken.

PR #613, opened 2021-05-03
Minor bug fix in ThingTalk utils

rayslxu:


gcampax (approved):
Argh, TypeScript fooled us!

PR #619, opened 2021-05-10
Implement stream based STT through almond-cloud WS

ad31c0:


gcampax (commented):
Some high-level questions and comments:
- Given we're touching the speech recognizer, it might be worthwhile to port it to TypeScript quickly without functional changes, and then change the interface. That way we can catch more errors statically.
- What happens if a platform doesn't have a VAD? Does it mean that all speech platforms need VAD? (if so, almond-gnome will need the same changes as almond-server, with the additional twist of a more complex build system)
- Do we still have the logic to recognize the command twice (first with a short silence and then with a longer silence) or no?

gcampax:
What's the length of silence that will end the command with this patch set? Did you test both issuing the command straight up and waiting for the beep, mabye half a second after the beep even?

ad31c0:
> What's the length of silence that will end the command with this patch set? Did you test both issuing the command straight up and waiting for the beep, mabye half a second after the beep even?

The length is now fixed to 96 consecutive silence frames, which is around ~1s of full silence, we may need to tune it better.
Consider also webrtcvad is not perfect in distinguishing between noise and voice, thus the consecutive silence counter is used to avoid false positives and improve accuracy.

lib/speech/speech_recognizer.ts
```diff
@@ -0,0 +1,233 @@
...
+    private _endDetected : boolean;
+    private _ended : boolean;
+    private _started : boolean;
+    private _vad : Tp.Capabilities.VadApi|null;
+    private consecutiveSilence : number;
+    private _bufferedMessages : Array<Buffer>;
+    private _piped : boolean;
+    private _endTimeout : NodeJS.Timeout|null = null;
+    private _connection : WebSocket|null = null;
+    private _listener : any;
```
gcampax, line 39: You can set `_listener` in the constructor and give it a function type instead of `any`

lib/speech/speech_recognizer.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    constructor(stream : Stream.Readable, initialBuffer : Buffer, vad : Tp.Capabilities.VadApi|null) {
+        // this is important, as the protocol does not allow chunks larger than 8192 bytes
```
gcampax, line 42: Is this still true for our protocol? Or should we let nodejs pick the best buffer size?

lib/speech/speech_recognizer.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    _write(chunk : Buffer, encoding : BufferEncoding, callback : (err ?: Error) => void) {
+        if (this._ended || this._endDetected) {
+            callback();
+            return;
+        }
+        //console.log('Sending chunk of length ' + chunk.length);
+        if (this._connection && this._connection.readyState === 1) {
+            // OPEN
+            if (this._vad && chunk.length === 320) {
```
gcampax, line 141: Why are we checking for an exact length here? node will merge and split buffers in the stream code

lib/speech/speech_recognizer.ts
```diff
@@ -0,0 +1,233 @@
...
+                if (this._vad.process(chunk))
+                    this.consecutiveSilence = 0;
+                else
+                    this.consecutiveSilence++;
+            }
+            if (this.consecutiveSilence === 96) {
+                this._endDetected = true;
+                console.log("VAD threshold reached", this.consecutiveSilence);
+                this._connection.send(undefined, {}, (err ?: Error) => callback(err));
```
gcampax, line 150: Are we allowed to send `undefined` on a WebSocket? What does that look like on the wire?

lib/speech/speech_recognizer.ts
```diff
@@ -0,0 +1,233 @@
...
+            });
+            connection.on('open', () => {
+                this._connection = connection;
+                console.log('STT connection opened');
+                this._connection.send(JSON.stringify({ ver: 1 }));
+                callback(connection);
+            });
+            connection.on('close', (code, reason) => {
+                if (code !== 1000) // 1000 = normal closure (eg timeout, or we closed on our side)
+                    console.log('Connection to STT service closed: ' + code + ' ' + reason);
```
gcampax, line 205: Are we actually setting the codes on the almond-cloud side?

lib/speech/speech_recognizer.ts
```diff
@@ -0,0 +1,233 @@
...
+                this.emit('error', e);
+            });
+        });
+    }
+
+    _ensureConnection() {
+        if (this._connection)
+            return Promise.resolve(this._connection);
+        else
+            return this._doConnect();
```
gcampax, line 219: Connections are not reusable in the new protocol, right? So this logic is not correct here.

lib/speech/speech_recognizer.ts
ad31c0: Yes. This causes the payload length field inside the WS packet header to be 0.

gcampax:
I think this PR is getting confused because you have _both_ speech_recognizer.js and speech_recognizer.ts. Please squash everything into two commits - one converting speech_recognizer to typescript and one with the new features (both buildable/testable, with a clean commit message) and force push to the PR branch.

gcampax:
You'll need to fix the merge conflicts or tests won't run.

gcampax:
There are still two files, the "port speech_recognizer to TS" commit needs to remove the old JS file.

gcampax (approved):
Alright, let's merge this when tests pass

PR #620, opened 2021-05-10
Small adjustments to FAQ

spacegaier:


gcampax:
Thanks for spotting this, and thanks for the patch!

PR #622, opened 2021-05-13
changing smartnews so that it plays audios one by one asynchronously …

rhulkb27:
…to allow for starting and stopping. Instead of queuing up all the audios at once, each audio is played one by one. Once it is done playing, it is saved so that if the user asks for news, he will not hear the same news again. This solves the following issue: https://github.com/stanford-oval/thingpedia-common-devices/issues/318 from thingpedia-common-devices.


gcampax:
I am not sure this is the best place to address this issue. The logic of not replaying existing news is SmartNews specific, it should be in the general Genie code. Even if we do put it here, the implementation is problematic: the global variable does not work if you have multiple Genie engines running in the same process (also, don't ever use `var` in modern JS), and the state should be saved somewhere persistently rather than just in memory. Otherwise the information will be lost when the assistant restarts.

rhulkb27:
Hi @gcampax, thanks for the feedback. I will try to fix the issues that you have outlined here. I do have a few questions though. I am not sure I understand what you mean by "general genie code", as the files I edited were in the genie-toolkit. As such, in which area of the code do you think it would be the best to change?

Additionally, how would you suggest I go about storing the news articles already read persistently?

gcampax:
Yeah, the genie-toolkit is what I'm referring to as "general genie code" - it's the code that is shared by all skills. Putting anything specific to one skill in here is not appropriate, although we can provide the mechanism here that skills can use to achieve a particular behavior.

As to where it should be placed, the reality is that there is no good place for this feature, which is why it hasn't happened yet.
So, let's make a strawman of the design, and let's see how far we get.

# Pause, Resume, Next
## Intended design
"pause", "resume", "next" commands, both by themselves and with the explicit "resume the news", have their intended behavior

## Implementation strategy
Given the audio controller is a core functionality, I think think "pause", "resume", "next", etc. could also be builtin commands that operate over the audio controller. Skills would register with the audio controller to receive pause/resume notifications like they do currently for "stop".
So we would have the following commands:
- "stop" (the exact word, no variations): special command that stops the current skill, whatever it is, and resets the full conversation
- "stop playing", "pause playing", "stop the sound", and other ambiguous variants: builtin function that calls into the AudioController, which then calls the current playing skill to stop/pause, or into the SpeechHandler to stop/pause the skills that use "audio" outputs (ie the news)
- "resume", "resume playing", and other ambiguous variants: builtin function that calls into the AudioController, which then calls the _previous_ playing skill to stop/pause, or into the SpeechHandler to resume whatever it was playing (ie. the news)
- "next", "previous", and other ambiguous variants: same as above: builtin function that calls into the AudioController
- "stop playing music", "resume playing music", "next song", and other _explicit_ commands about music: regular Thingpedia function that calls into Spotify or Home Assistant to stop/pause/next
- "stop playing news", "resume playing news", "next article", and other _explicit_ commands about news: regular Thingpedia function in SmartNews that calls into the AudioController with the SmartNews skill as a parameter; the AudioController checks if the currently playing skill is SmartNews; is so, performs the action, if not, nothing happens.

(Tagging @almakantara here because this change also affects "pause", "resume", "next" of Home Assistant media players)

# Tracking the current article (aka "tell me more about this")
## Intended design
While news summaries are playing, the user says "tell me more about this". The summaries stop playing, and the assistant starts reading the full article of whatever the summary was when the user gave the command.

A design challenge, is that summaries are relatively short, so by the time the user has said "tell me more about this", the next news article might already be playing, and the user might be off. We might need to track the article _at the time the user starts the command_, or give up entirely on the feature and require the user to search the news by voice.

## Implementation strategy
The concept of "this" in natural language is one of the trickiest, because it's pretty much always ambiguous and context dependent. [This paper](https://jackieyang.me/files/dothishere.pdf) (by my colleague Jackie Yang) describes how to talk about "this" in the context of a mobile UI. It worked by creating a new ThingTalk value, with syntax `$context.selection`. We could presumably do the same: create a special value `$context.current_audio` or similar that refers to the audio result currently being played (or better, the ID of the item associated with the audio result currently being played - a bit tricky to plumb through but it simplifies the skills). Then SmartNews would have a function that takes the ID of a news item and returns the full body and audio URL of the full article, and we're done.
(We'll need changes to the synthesis to support this design. Not easy to do.)

# Tracking of previously read articles
## Intended design
The user says "get the news". 10 articles are returned, shown in the web UI and played in voice. Next time the user says "get the news", 10 different articles are returned.

TBH, I'm not 100% sure we actually want this feature. It was discussed in some design meetings but the more I think about it it's weird that "get the latest news" becomes progressively older and older as you run out of more recent news.

## Implementation strategy
There are a couple of options. I think the one that comes closest to the right design is that CardFormatter associates a device instance (Tp BaseDevice) with every media result object it produces. The SpeechHandler would look at the device instance, check for some new extended interface, and call into it to record that the news article was played. The device instance would store the information in some persistent storage (to be designed -- see also #617), and use that persistent storage in the `get_article` function to filter out anything that was already played.

# Overall strategy
These are not small features. I think we should:
- Pick one feature at a time
- Flesh out the design
- Divide it in subtask
- Implement each subtask

gcampax:
Media control commands were implemented. The speech code in genie-toolkit is on its way out, so the rest needs to be handled differently. Probably it will be handled by not making news play audio.

PR #624, opened 2021-05-18
add postgres db client for cockroachdb

jgd5:
Ran  ./node_modules/.bin/ts-node -T test/engine/index.js with a local cockraochdb.

A couple remaining issues requiring your inputs:

* What is the best way to configure which database implementation to use?
* How do we plan to run tests?   Download the cockroach binary (it's about 200MB) on the fly or preconfigure cockroachdb on test machine.

### DIFF
--- a/lib/engine/db/postgres/index.ts
+++ b/lib/engine/db/postgres/index.ts
@@ -1,0 +1,150 @@
+// -*- mode: typescript; indent-tabs-mode: nil; js-basic-offset: 4 -*-
+
+//
+// This file is part of Genie
+//
+// Copyright 2021 The Board of Trustees of the Leland Stanford Junior University
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import * as Tp from 'thingpedia';
+
+import type * as DB from '..';
+
+import { Pool, PoolClient, } from 'pg';
+
+import LocalTable from './local_table';
+import SyncTable from './sync_table';
+
+
+export interface PostgresPlatform extends Tp.BasePlatform {
+    getPostgresConfig() : any;
+}
+
+export function query(client : PoolClient, string : string, args : unknown[]) {
+    return new Promise<void>((resolve, reject) => {
+        client.query(string, args, (err) => {
+            if (err)
+                reject(err);
+            else
+                resolve();
+        });
+    });
+}
+
+export function selectAll(client : PoolClient, string : string, args : unknown[]) {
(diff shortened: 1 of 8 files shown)


lib/engine/db/postgres/schema.sql
```diff
@@ -0,0 +1,145 @@
...
+--    cockroach start-single-node  --insecure --listen-addr=localhost:8888 --store=type=mem,size=640MiB
+--     
+-- The following command can be use to initialize the databse for testing:
+--    cockroach sql --url postgresql://root@localhost:8888?sslmode=disable < schema.sql 
+
+drop database if exists thingengine;
+create database thingengine; 
+use thingengine;
+
+drop table if exists app ;
```
gcampax, line 15: The only tables we need are `app`, `channel`, `device`, `device_journal`. The rest can go.

lib/engine/db/postgres/index.ts
```diff
@@ -0,0 +1,150 @@
...
+                reject(err);
+            else
+                resolve();
+        });
+    });
+}
+
+export function selectAll(client : PoolClient, string : string, args : unknown[]) {
+    return new Promise<any[]>((resolve, reject) => {
+        client.query(string, args, (err, res) => {
```
gcampax, line 48: I swear I originally modeled the `query`, `selectAll`, `selectOne` APIs based on postgres (many years ago before deciding on mysql for almond-cloud instead). Maybe that was a different postgres library?
Anyway interesting that we end up wrapping anyway.

lib/engine/db/postgres/index.ts
```diff
@@ -0,0 +1,150 @@
...
+                });
+                this._transactionQueue.set(client, newqueue);
+            });
+        });
+    }
+
+    ensureSchema() {
+        // Database is initialized out of band
+        // eslint-disable-next-line no-promise-executor-return
+        return new Promise<void>((resolve, reject) => resolve());
```
gcampax, line 140:
```suggestion
        return Promise.resolve();
```
(alternatively, make the `ensureSchema()` function async and leave it empty)

lib/engine/db/postgres/index.ts
```diff
@@ -0,0 +1,150 @@
...
+                }
+                resolve(callback(c));
+                release();
+            });
+        });
+    }
+
+    withTransaction<T>(transaction : (client : PoolClient) => Promise<T>) : Promise<T> {
+        return this.withClient((client) => {
+            const queue = this._transactionQueue.get(client) || Promise.resolve();
```
gcampax, line 115: We definitely don't need the transaction queue here, this was a hack to fix broken sqlite's locking. With a real DB, we have real locking and can run multiple transactions concurrently...

gcampax (commented):
This looks ok, but i wonder what's the best approach for modularity: should the whole postgresdb implementation live in almond-cloud (where we can deal with schema changes centrally)? Then you'd call [`platform.getCapability`](https://almond-static.stanford.edu/jsdoc/thingpedia/latest/classes/baseplatform.html#getcapability) and get back the DB interface directly.
Partially also because I want to keep the dependencies of genie-toolkit somewhat lean, given it goes into all sorts of docker images.

Alternatively, a generic `platform.getDBConfig()` that returns an object, or a `platform.getDatabaseUrl()`, and we use the scheme to distinguish the backend.

Also some comments on the implementation.

jgd5:
Abandon this PR in favor of #648

PR #626, opened 2021-05-20
Update the pruning size and depth default values

gcampax:
For generate and generate-dialogs.

Fixes an issue with the basic tutorial, which doesn't pass the command-line arguments and runs out of memory very quickly.


gcampax:
Uh. I checked, I use 9 for multiwoz, and 8 for thingpedia. Maybe 8 should be the default?

rayslxu:
> Uh. I checked, I use 9 for multiwoz, and 8 for thingpedia. Maybe 8 should be the default?

as long as it runs on 8gb ram machine, it's fine

PR #632, opened 2021-05-23
Listen audio if we expect some input from the user

ad31c0:


gcampax (commented):
I don't think this is right. We should run the VAD and start listening when voice is detected. Otherwise there will be silence if the user waits a bit to reply, and we'll miss the beginning if the user starts speaking too early.

gcampax:
Ah also: if you wait synchronously inside setExpected, you will delay processing of commands inside the conversation as well, and potentially delay delivering messages to the web interface. So we probably shouldn't do that.

gcampax:
The speech code in genie-toolkit is on its way out. Closing.

Issue #590, opened 2021-04-28
Does not compile with npm 7


gcampax:
For some unscrutable reason, npm 7 changed how git deps are prepared, and now repos that use genie-toolkit (like almond-server or thingpedia-common-devices) no longer build genie-toolkit correctly: the templates fail to build because they fail to locate the genie-toolkit.d.ts file

I think the solution here will be to remove entirely the ability to specify templates outside of Genie. Templates will be shipped with Genie and there will only ever be one template pack (+ flags). Programmability of dialogues will come in a different way, by configuring the transaction/slot filling templates with custom ThingTalk logic (see https://wiki.almond.stanford.edu/en/genie/design-notes/unified-control-language for a summary of current designs).
@rayslxu opinions? We'll drop sempre templates when we do that.

(originally reported independently by @aydanpirani and @james-zhuang)

rayslxu:
Agree with shipping pre-compiled templates.

PR #636, opened 2021-05-28
Clean up the template build system

gcampax:
This RFC PR comes from the observation that people won't be writing Genie templates and distribute them as separate packages. 

Core features that require new ThingTalk will be developed inside Genie itself (or maybe as a fork of Genie), while custom dialogues will be represented in some higher level way that doesn't involve writing templates directly.

Once we accept that, we can clean up how templates and the rest of the sentence generator runtime are linked together. This is a cleanup in preparation for changing how agent policies are expressed, which will change all the business around context tags and "context"/"answer" functions in the dialogue templates, and couple dialogue templates even tighter to the rest of the dialogue agent.

Side effects of this PR:
- custom template files are gone, and the --template option to `genie generate` is no longer useful: it is still there to select betwene "basic" (exact match) and "single-sentence" templates (full thingtalk)
- #590 is fixed because we don't call typescript twice


rayslxu (approved):
on one hand, I like the changes with a much cleaner structure, but on the other hand, I feel I will have a painful time when rebasing my ongoing branches...

gcampax:
Do you have a lot of outstanding branches? I'm hoping that git recognizes these are just renames and will do the right thing, but you might have to cherry-pick changes by hand...

gcampax:
It actually looks like git can pick up the rename just fine, I am seeing basically no merge conflicts when I merged with the other PR on the command-line.

rayslxu:
> It actually looks like git can pick up the rename just fine, I am seeing basically no merge conflicts when I merged with the other PR on the command-line.

oh, cool. i have just one branch for wikidata now actually and it's not touching the templates so far. so it should be fine anyway

PR #638, opened 2021-05-31
Cherry-pick more work from Covid Genie branch

gcampax:
In particular, the refactoring of the dialogue loop into a Chirpy-like architecture with multiple potential handlers for a command, called in parallel.


languages/thingtalk/load-thingpedia.ts
```diff
@@ -127,6 +127,8 @@ export default class ThingpediaLoader {
     private _errorMessages : Map<string, Record<string, ParsedPlaceholderPhrase[]>>;
     private _resultPhrases : Map<string, NormalizedResultPhraseList>;
 
+    private _initialFunction : Ast.FunctionDef|null;
```
rayslxu, line 142: I feel like we would like this to be an array in the future

languages/thingtalk/load-thingpedia.ts
```diff
@@ -1192,10 +1209,32 @@ export default class ThingpediaLoader {
             }
             this._initialFunction = functionDef;
         }
+
+        const followUp = functionDef.getImplementationAnnotation<string>('follow_up');
+        if (followUp) {
```
rayslxu, line 1214: when I saw the if statement in the annotation, I was like "this is neat, how is it done in thingtalk?", then I saw this... lol
we do plan to generalize it and do it in thingtalk in the future right?

rayslxu (approved):
look good

languages/thingtalk/load-thingpedia.ts
gcampax: The plan is not to use the `#[follow_up]` annotation at all. Instead, if you want custom agent behavior you write a tiny JS or TT function that replaces the agent policy entirely, while the skill is active. The design is still in flux, between me Monica and Chaofei. You can join the discussion on Slackl if you're curious.

PR #648, opened 2021-06-03
Add database proxy

jgd5:
This proxy implements AbstractDatabase and proxys all requests to a cloud database.

Ran ./node_modules/.bin/ts-node -T test/engine/index.js  with database proxy enabled.

Cloud sync seems to be broken. Sever side seems returning bad responses. So I commented it out for now.

### DIFF
--- a/lib/engine/db/dbproxy/local_table.ts
+++ b/lib/engine/db/dbproxy/local_table.ts
@@ -29,52 +29,29 @@ export default class LocalTable<RowType> {
         this._userId = userId;
     }
 
-    getAll() : Promise<RowType[]> {
-        return new Promise<any[]>((resolve, reject) => {
-            Tp.Helpers.Http.get(`${this._baseUrl}/localtable/user_${this.name}/${this._userId}`)
-            .then( (resp : string) => { 
-                resolve(JSON.parse(resp)['data']);
-            }).catch( (err) =>{
-                reject(err); 
-            });
-        });
+    async getAll() : Promise<RowType[]> {
+        const resp = await Tp.Helpers.Http.get(`${this._baseUrl}/localtable/user_${this.name}/${this._userId}`);
+        return JSON.parse(resp)['data'];
     }
 
-    getOne(uniqueId : string) : Promise<RowType|undefined> {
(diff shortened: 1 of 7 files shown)


gcampax:
Are you sure cloud sync is broken? It was working until yesterday

jgd5:
> Are you sure cloud sync is broken? It was working until yesterday

Can you give it a try? It fails with the same error with master (using sqlitedb).

lib/engine/db/dbproxy/local_table.ts
```diff
@@ -0,0 +1,80 @@
...
+    getAll() : Promise<RowType[]> {
+        return new Promise<any[]>((resolve, reject) => {
+            Tp.Helpers.Http.get(`${this._baseUrl}/localtable/user_${this.name}/${this._userId}`)
+            .then( (resp : string) => { 
+                resolve(JSON.parse(resp)['data']);
+            }).catch( (err) =>{
+                reject(err); 
+            });
+        });
+    }
```
gcampax, line 41: The whole function can be written succintly as:
```
    async getAll() : Promise<RowType[]> {
        const response = await Tp.Helpers.Http.get(`${this._baseUrl}/localtable/user_${this.name}/${this._userId}`);
        return JSON.parse(resp)['data']);
    }
```
(same for the other functions).

Generally speaking, you only need to use the Promise constructor when dealing with legacy APIs that are not themselves based on promises. You don't need it with modern APIs.
Also, you should avoid the TypeScript type `any` if there is no reason for it.

gcampax (commented):
I'm not seeing any obvious error, although the difference between Genie/Almond is suspicious. Maybe rebase the PR on top of master?

gcampax (approved):
Yeah this looks good! I suppose we're still missing authentication. Do you want to add it right away in this PR? I'm assuming you'll be passing a token in DatabaseProxyConfig and passing an Authorization header.

jgd5:
I plan to add authentication after we iron out the whole security model.  Maybe this something Emily can help.  Right now, I hope to get the k8s backend to work first.

PR #649, opened 2021-06-04
Log conversation data

kevintangzero:


lib/dialogue-agent/conversation.ts
```diff
@@ -571,24 +570,10 @@ export default class Conversation extends events.EventEmitter {
...
-
-        serializer.pipe(output);
+        const db = this.engine.db.getLocalTable('conversation');
         for (const dialogue of this._log)
-            serializer.write({ id: this.id, turns: dialogue.turns.map((log) => log.turn) });
-        serializer.end();
-
-        await StreamUtils.waitFinish(output);
+            for (const turn of dialogue.turns)
+                await db.insertOne(this.id, turn.turn);
```
gcampax, line 576: this.id is the ID of the conversation, which is the same for each turn, and also in practice always the same for all dialogues. You need to make up a unique ID for each turn.

Also, I think we should write each turn to the database incrementally when the turn is added to the log if recording is enabled, rather than in one fell swoop when saveLog is called. This is more robust in case of errors. 

Also, now that the log file no longer exists, the code immediately after (which is used to show the log to the user for error reporting) is no longer correct. You'll need to add a method to read the log back from the database.

lib/engine/db/sqlite/schema.sql
```diff
@@ -31,3 +31,15 @@ create table channel (
...
+
+create table conversation (
+       uniqueId varchar(255) primary key,
+       context text default null,
+       agent text default null,
+       agent_target text default null,
+       intermediate_context text default null,
+       user text default null,
+       user_target text default null
+);
```
gcampax, line 45: You'll want to add the migration code in db/sqlite/migrations. Add a function `migrateTo9` to the MIGRATION array that will create the table.

gcampax:
Btw, the error in the dialogue agent test is because the test scaffold in test/agent doesn't mock engine.db, so you'll have to add that.

kevintangzero:
Updated:
1. Use uuid
2. Insert data incrementally in `turnFinished()`
3. Add migrate code

gcampax (approved):
Looks good!

gcampax:
Oh and fix lint please!

gcampax:
Oh one thing I didn't notice: we should probably put the conversation ID, the timestamp, and the previous turn ID (or null) in the database, otherwise it's not possible to reconstruct the conversation log from the turns alone.

kevintangzero:
Updated:
1. Add conversation id, previous turn id, dialogue id, and timestamp
2. Fix lint/tab problems

kevintangzero:
Add mock db and local table for testing

lib/engine/db/index.ts
```diff
@@ -45,9 +45,25 @@ export interface DeviceRow {
...
+    dialogueId : string;
+    context : string|null;
+    agent : string|null;
+    agentTimestamp : string|null;
+    agentTarget : string|null;
+    intermediateContext : string|null;
+    user : string;
+    userTimestamp : string|null;
+    userTarget : string;
+}
```
gcampax, line 61: I think we need to store the vote and comment from the user here too.

gcampax (approved):
Overall looks good.
It's a bit weird that we're still also storing the logs on disk, and it's worth making a cleanup PR after this.

kevintangzero:
Added vote and comment fields in the database.

PR #651, opened 2021-06-08
Tuning the new VAD implementation

gcampax:


almakantara:
After some test with different VAD Threshold, the actual value of 96 seems too short. A good value seems to be 256

with the sentence (at a normal speech speed) "Help me to start the vacuum cleaner" these are some results:

VAD threshold reached 96
Recognized as "Computer help me to start the."

VAD threshold reached 96
Recognized as "Computer help me to start the vacuum."

VAD threshold reached 96
Recognized as "Computer help me to start."

VAD threshold reached 128
Recognized as "Computer help me to start the vacuum."

VAD threshold reached 128
Recognized as "Computer.Help me to start the."

VAD threshold reached 128
Recognized as "Computer.Start the vacuum clean."


VAD threshold reached 256
Recognized as "Computer.Help me to start the vacuum cleaner."

gcampax:
This is quite surprising. I found the threshold of 96 to be way to long, and I'm proposing to lower it to 20 (0.2 seconds).
256 would be 2.5 seconds which is a long pause, especially when compounded with extra noise that can reset the counter.

gcampax:
Ok, @almakantara has tested that a threshold of 20 works, are we good to merge this?

PR #655, opened 2021-06-10
remove bad chars [,.] from the utterance

ad31c0:


gcampax:
We don't want to unconditionally remove all characters - some are actually meaningful (think decimal separators), and we don't want to replace with nothing cause that would merge two consecutive words. We need to be more careful with this.

gcampax:
I think this is still too aggressive. I think we really only want to catch the case of a period that is added as an artifact by MS STT in between two words without spaces. And you don't need to clean up spaces, later tokenization of the sentence will take care of that.

Also, I feel like this is such an artifact of the MS backend that I would do it in the cloud speech API, not here, in a random place in the middle of Genie.

Issue #656, opened 2021-06-14
Cloud sync failing on the cloud side


gcampax:
```
(node:441) UnhandledPromiseRejectionWarning: TypeError: undefined is not a function
    at /opt/almond-cloud/node_modules/genie-toolkit/dist/lib/engine/db/syncdb.js:255:62
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
(node:441) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 9)
HTTP request failed: 404 page not found
Processing syncdb change for device failed HTTPError: Unexpected HTTP error 404 in request to http://dbproxy.staging.svc.cluster.local:8200/synctable/user_device/623/io.home-assistant%2Fhttp%3A%2F%2F127.0.0.1%3A8123/1623713798444
    at IncomingMessage.<anonymous> (/opt/almond-cloud/node_modules/thingpedia/dist/helpers/http.js:139:35)
    at IncomingMessage.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  code: 404,
  url: 'http://dbproxy.staging.svc.cluster.local:8200/synctable/user_device/623/io.home-assistant%2Fhttp%3A%2F%2F127.0.0.1%3A8123/1623713798444',
  detail: '404 page not found'
}
```

This is causing Travis to hang because the connection is dropped incorrectly by the cloud.

gcampax:
On a different user, I'm seeing this error:
```
(node:343) UnhandledPromiseRejectionWarning: TypeError: undefined is not a function
    at /opt/almond-cloud/node_modules/genie-toolkit/dist/lib/engine/db/syncdb.js:255:62
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[...]

PR #658, opened 2021-06-15
Fix dbproxy sync interface

gcampax:
Use a struct based interface instead of a tuple interface because
it's more type-safe and also it works better from Go

Should help with #656


jgd5 (approved):
Thanks for the fix.

PR #659, opened 2021-06-15
Update thingtalk dependency

emilyjchang:


gcampax:
When you edit package.json you also need to run `npm install` to update package-lock.json, and commit that too, otherwise the change won't have any effect.

gcampax (approved):
Anyway good to merge when Travis is happy...

PR #660, opened 2021-06-15
Changes to enable confidence-based dispatch

gcampax:
So we can actually dispatch to third-party services if necessary.


s-jse (commented):
What is the use for a `score`, if we have the three-part intent?

gcampax:
They map to different confidence estimators, and they have different purposes - one is to detect potential errors, and one is for OOD detection.

gcampax:
Note that the `intent` is really two part, the `intent.ignore` value is always 0 and only set for API compatibility.

s-jse (approved):

PR #662, opened 2021-06-17
Refactor logging to avoid files

gcampax:
This is a follow up to #649 that cleans up the conversation logging code: now that we store everything in the main database, we don't need to write to local files, and we don't need to keep track of dialogues in memory.

Instead, we can store everything into the database and reconstruct the sequence of turns dynamically when needed.

Note that the reconstructed log is essentially identical to what we had before (see test changes).


rayslxu (approved):
looks good

PR #667, opened 2021-06-22
Add support for authentication in dbproxy interface

gcampax:
Related to https://github.com/stanford-oval/almond-cloud/pull/1020

### DIFF
--- a/lib/engine/db/dbproxy/local_table.ts
+++ b/lib/engine/db/dbproxy/local_table.ts
@@ -21,22 +21,22 @@ import * as Tp from 'thingpedia';
 export default class LocalTable<RowType> {
     name : string;
     private _baseUrl : string;
-    private _userId : number;
+    private _auth : string|undefined;
 
-    constructor(name : string, baseUrl : string, userId : number) {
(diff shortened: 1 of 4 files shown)


lib/engine/db/dbproxy/local_table.ts
```diff
@@ -21,22 +21,22 @@ import * as Tp from 'thingpedia';
 export default class LocalTable<RowType> {
     name : string;
     private _baseUrl : string;
-    private _userId : number;
+    private _auth : string|undefined;
 
-    constructor(name : string, baseUrl : string, userId : number) {
+    constructor(name : string, baseUrl : string, accessToken : string|undefined) {
```
jgd5, line 26: should accessToken be required since dbproxy api will be broken without a user id ?

jgd5 (approved):
LGTM

lib/engine/db/dbproxy/local_table.ts
gcampax: I wanted to keep the interface a bit more generic in case we change how dbproxy works in the future. The contract right now is that the platform decides if the access token is required or not. If the access token is provided, it is also required by the API.

PR #683, opened 2021-07-04
Add thingpedia-dir cmdline argument to linter

ryachen01:
Allows linter to load manifests from a local directory


gcampax (approved):
Yeah

PR #692, opened 2021-07-14
Two changes to reduce the sequence length passed to genienlp

gcampax:
The recently approved movie skill returns the full list of actors to all movies it returns. This can be quite a long list, and it's too long to encode to BART, even if every actor is a single `GENERIC_ENTITY_org.themoviedb:actor_<n>` token (which is split into `ENTITY actor <n>` before BART tokenization).

To avoid that, we reduce the number of history items encoded in the context (which also helps the model by giving a stronger recency bias) and we cut the length of array values in the result.

This was caught in the staging environment which keeps going down (https://github.com/stanford-oval/genienlp/issues/174)


s-jse (approved):
Looks good to me.
Would be interesting to see the difference (if any) this makes in terms of accuracy as well.

PR #716, opened 2021-07-19
Change dialogue templates to be more command-oriented

gcampax:
1. Follow-ups are treated as carrying the context only if they
   use words like "anything", "something" (so they definitely
   depend on context)
2. Every command where the agent is not asking a question becomes
   an occasion to start a brand new command.
3. Refinement proposals from the agent are disabled.

The result is that commands that don't depend on context should
become more robust as the model is basically never trained to
carry over context if the command looks like a full command.
Yet, true follow-ups, including questions about the current results,
expressions that use "this", and search refinements using "anything",
are still recognized.

These changes are gated behind a "multiwoz" flag. If the flag is
set, the previous behavior remains. If the flag is not set, the
new templates are enabled.

----

@monicalam this is the "nuclear option" of disabiling dialogues without losing any features.


s-jse (approved):
Should `dlgthingtalk` and `error-correction` Makefiles in genie-workdirs be updated as well?

gcampax:
Yeah they should. I will update them, thanks for the reminder. I'm going to test this in Kubeflow before merging.

PR #724, opened 2021-07-29
Add activity monitor to detect engine idleness

jgd5:


lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    _startQuiesce() {
+        console.log(`Activity monitor starts quiesce timer ${this._quiesceTimeoutMillis} ms`);
+        this._quiesceTimeout = setTimeout(() => {
+            this._quiesceTimeout = null;
+            console.log('Activity monitor emits idle event');
+            this.emit("idle");
+        }, this._quiesceTimeoutMillis);
+        for (const [_, conversation] of this._assistant.getConversations())
+            conversation.sendPing();
```
gcampax, line 102: This might be marginally cleaner with a `pingAll` method on AssistantDispatcher. It's cosmetic though.

lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+    constructor(appdb : AppDatabase, assistant : AssistantDispatcher, options : ActivityMonitorOptions) {
+        super();
+        this._running = false;
+        this._appdb = appdb;
+        this._assistant = assistant;
+        this._lastUpdate = 0;
+        this._idleTimeout = null;
+        this._idleTimeoutMillis = options.idleTimeoutMillis || DEFAULT_IDLE_TIMEOUT;
+        this._quiesceTimeout = null;
+        this._quiesceTimeoutMillis = options.quiesceTimeoutMillis || DEFAULT_QUIESCE_TIMEOUT;
```
gcampax, line 55: We should probably add a comment explaining the sequence of timers, in case we forget what the logic should be here.

Also, is 0 a valid value here? (If so, we should use `??` instead of `||`)

lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+            return;
+        if (this._quiesceTimeout !== null) {
+            clearTimeout(this._quiesceTimeout);
+            this._quiesceTimeout = null;
+        }
+        this._maybeStartActivityTimeout();
+    }
+
+    _maybeStartActivityTimeout() {
+        if (this._appdb.getAllApps().length > 0)
```
gcampax, line 80: This is a bit unnecessarily expensive because it converts the internal map to an array. We can add a method to AppDatabase to check empty/not-empty.

lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+    private _running : boolean;
+    private _appdb : AppDatabase;
+    private _assistant : AssistantDispatcher;
+    private _lastUpdate : number;
+    private _idleTimeout : NodeJS.Timeout|null;
+    private _idleTimeoutMillis : number;
+    private _quiesceTimeout : NodeJS.Timeout|null;
+    private _quiesceTimeoutMillis : number;
+
+    constructor(appdb : AppDatabase, assistant : AssistantDispatcher, options : ActivityMonitorOptions) {
```
gcampax, line 46: These days, I've taken to putting options interfaces inline as:
```
constructor(appdb : AppDatabase, assistant : AssistantDispatcher, options : {
   idleTimeoutMillis : number;
   quiesceTimeoutMillis : number
}) {
```
It looks a bit nicer in the generated documentation because you can document each option with `@param options.idleTimeoutMillis` etc.

Also, the options should be `?: number` cause they don't need to be specified.

lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+            conversation.sendPing();
+    }
+
+    stop() {
+        this._running = false;
+        if (this._idleTimeout !== null) 
+            clearTimeout(this._idleTimeout);
+        if (this._quiesceTimeout !== null)
+            clearTimeout(this._quiesceTimeout);
+        console.log('Activity monitor stopped');
```
gcampax, line 111: Technically, you should disconnect the events here too. Not that it matters because the whole engine is going down and the appdatabase won't be emitting more signals, but it's good practice and it helps with leaks.

lib/engine/index.ts
```diff
@@ -337,6 +367,8 @@ export default class AssistantEngine extends Tp.BaseEngine {
      */
     run() : Promise<void> {
         this._running = true;
+        if (this._activityMonitor)
+            this._activityMonitor.start();
```
gcampax, line 371: You can add _activityMonitory to the list of engine modules and start/stop will be called when the engine initializes (in open) and deinitializes (in close).

lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+        if (this._idleTimeout !== null)
+            return;
+        if (this._quiesceTimeout !== null) {
+            clearTimeout(this._quiesceTimeout);
+            this._quiesceTimeout = null;
+        }
+        this._maybeStartActivityTimeout();
+    }
+
+    _maybeStartActivityTimeout() {
```
gcampax, line 79: This should be `private`

lib/engine/activity_monitor.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export class ActivityMonitor extends events.EventEmitter {
+    private _running : boolean;
+    private _appdb : AppDatabase;
+    private _assistant : AssistantDispatcher;
+    private _lastUpdate : number;
+    private _idleTimeout : NodeJS.Timeout|null;
+    private _idleTimeoutMillis : number;
+    private _quiesceTimeout : NodeJS.Timeout|null;
+    private _quiesceTimeoutMillis : number;
```
gcampax, line 44: Given that the k8s controller polls the worker to see if engines are idle, should we also add a public property/accessor to return the idle status of the engine? That way the worker doesn't need to connect to the idle signal really.

test/engine/test_apps.js
```diff
@@ -32,6 +32,24 @@ async function collectOutputs(app) {
...
+async function testActivityMonitor(engine) {
+    const app = await engine.createApp('timer(base=$now,interval=10s) => @org.thingpedia.builtin.test.get_data(count=2, size=10byte) => notify;',
+            { icon: 'org.foo', uniqueId: 'uuid-timer-foo', name: 'some app', description: 'some app description' });
+    let idle = false;
+    engine.activityMonitor.on('idle', () => idle = true);
+
+    // inactivity timers should not start when there are apps
+    await delay(2000);
+    assert(engine.apps.hasApp(app.uniqueId));
+    assert(idle === false);
```
gcampax, line 44: `assert.strictEqual(idle, false)` will give you better errors when the assertion fails.

gcampax (approved):
Overall looks good. Thanks for adding tests as well.
I have some suggestions and some comments, if you like.

lib/engine/activity_monitor.ts
```diff
@@ -24,47 +24,71 @@ import AssistantDispatcher from '../dialogue-agent/assistant_dispatcher';
...
-export interface ActivityMonitorOptions  { 
-        idleTimeoutMillis : number;
-        quiesceTimeoutMillis : number;
-}
 
+export enum ActivityMonitorStatus {
+    Starting = "starting",
+    Running = "running",
+    Idle = "idle",
+    Stopping = 'stopping'
```
gcampax, line 32: Nice idea to use an enum, but by convention, I tried to use UPPER_CASE for enums in typescript

PR #738, opened 2021-08-02
Add script to generate translated template to work on

Krish-sysadmin:
This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po. 

Edit the language variable in translator.py to change the language that it translates too. Using the https://huggingface.co/facebook/m2m100_418M model for translation from FaceBook so can't get rate limited for the translations.

### DIFF
--- a/po/perform_translations/README.md
+++ b/po/perform_translations/README.md
@@ -33,9 +33,11 @@ After this you can examine and manually make required final changes to the newfi
 
 
 ```
-
-mv newfile.po fr.po
+msmerge newfile.po fr.po --output-file= fr.po
 
 ```
 
 Replace fr.po with $language.po . Now you have your translated po file.
+
+
+
--- a/po/perform_translations/README.md
+++ b/po/perform_translations/README.md
@@ -1,4 +1,4 @@
-This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po.  In order to use this translator template generator you will want to run this to install the "polib" and "dl-translate" modules.
+This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po.  In order to use this translator template generator you will want to run this to install the "polib" and "dl-translate" modules. Make sure GNU gettext utilities are installed!
 
 
 ```
--- a/po/perform_translations/README.md
+++ b/po/perform_translations/README.md
@@ -1,0 +1,14 @@
+This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po.  In order to use this translator template generator you will want to run this to install the "polib" and "dl-translate" modules.
+
+
+```
+pip3 install polib dl-translate
+
+```
+
+
+Further documentation on the modules is available here:
+
+- [polib](https://pypi.org/project/polib/)
+- [dl-translate](https://pypi.org/project/dl-translate/)
+
--- a/po/perform_translations/README.md
+++ b/po/perform_translations/README.md
@@ -1,5 +1,4 @@
-This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po.  In order to use this translator template generator you will want to run this to install the "polib" and "dl-translate" modules. Make sure GNU gettext utilities are installed!
-
+This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po.  In order to use this translator template generator you will want to run this to install the "polib" and "dl-translate" modules. Make sure GNU gettext utilities are installed! In addition, make sure you are running Python 3.5 or above as subprocess.run() function was added in Python 3.5.
 
 ```
 pip3 install polib dl-translate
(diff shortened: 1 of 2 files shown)


po/perform_translations/translator.py
```diff
@@ -0,0 +1,59 @@
+import re
+import dl_translate as dlt
+import polib
+
+# Variables for translator to edit
+
+language = "fr" # change to desired language to convert TO
```
gcampax, line 7: This should be a parameter in the command-line, not hardcoded.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,59 @@
...
+            r"\$\w+",  # $link
+            r"[{|}]",  # {a | b}
+        )
+    )
+    parts = re.split(reg, messageId)
+    placeholders = re.findall(reg, messageId)
+    translation = []
+    for part in parts:
+      if part.strip() != "":
+         translation.append(mt.translate(part, source=dlt.lang.ENGLISH, target=dlt.lang.FRENCH))
```
gcampax, line 35: Target should be the same as language.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,59 @@
...
+for messageId in msgIdArray:
+    reg = "|".join(
+        (
+            r"<\${\w+?}>",  # <${link}>
+            r"\${\w+?}",  # ${link}
+            r"<\$\w+?>",  # <$link>
+            r"\$\w+",  # $link
+            r"[{|}]",  # {a | b}
+        )
+    )
```
gcampax, line 29: This regular expression is not sufficient to cover all the syntax of template strings. The full syntax is at https://github.com/stanford-oval/genie-toolkit/blob/master/lib/utils/template-string/grammar.pegjs

In particular, I'm not sure how you're handling grammar flags, and phrases inside a ${select} clause.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,59 @@
...
+         translation.append(part)
+
+    placeholder_i = iter(placeholders)
+    res = ""
+    for part in translation:
+        res += part + " "
+        res += next(placeholder_i, "") + " "
+
+    a = re.sub(' +',' ', res)
+    arrayOfTranslated.append("msgstr " + f'"{a}"')
```
gcampax, line 46: This is not correct in case the translation includes a quote mark or newline. You should use polib to generate the file.

gcampax (changes requested):
Thank you for writing this script. Some changes are needed before we can merge it.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,97 @@
...
+mt = dlt.TranslationModel()
+
+
+# Variables for translator to edit
+
+print(mt.get_lang_code_map()) 
+
+language = str(input("What is your desired language to translate to? Input the language code: ")) # change to desired language to convert TO
+Reportmsgid = str(input("What is your desired Report Msgid Bugs To? " ))
+POT_Creation_Date = str(input("What is your desired POT Creation Date? "))
```
gcampax, line 15: You don't need this, you'll fetch this from the POT.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,97 @@
...
+PO_Revision_Date = str(input("What is your desired PO Revision Date? "))
```
gcampax, line 16: This should be the current time.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,97 @@
...
+Last_Translator = str(input("Who do you want as the last translator? "))
+Language_Team = str(input("What is the language team you want the translation to be in? Example: English <myteam@example.com> "))
+
+
+while language not in mt.available_codes():
+  print("Sorry try again! The language code is incorrect")
```
gcampax, line 22: You should maybe say that the language is not supported (not all valid languages are supported by dl-translate)

po/perform_translations/translator.py
```diff
@@ -0,0 +1,97 @@
...
+for i in theActualArray:
+  entry = polib.POEntry(
+      msgid=i[0],
+      msgstr=i[1]
+  )
+  po.append(entry)
+
+
+
+os.system('msgmerge newfile.po' +  language + '.po' + '> newfile.po')
```
gcampax, line 93: As a rule, you never want to use os.system, because it introduces a risk of shell injection vulnerability. Use the subprocess package instead.

po/perform_translations/README.md
```diff
@@ -0,0 +1,14 @@
+This python script helps generate a translated template to work on to speed up the translation workflow. It generates a file $language.po.new that you can fix up and then rename to $language.po.  In order to use this translator template generator you will want to run this to install the "polib" and "dl-translate" modules. Make sure GNU gettext utilities are installed!
+
+
+```
+pip3 install polib dl-translate
+
+```
+
+
```
gcampax, line 9: You might want to give an example of how to actually call the script? Ie, what commands you need to run before the script, from what directory you run it, with what options, etc.

po/perform_translations/translator.py
Krish-sysadmin: Fixed this @gcampax

po/perform_translations/translator.py
gcampax: Ok yeah this makes sense. I suppose people will find a mix of English and target language and then deal with it. At least the see the original and can make sense of it.

Krish-sysadmin: Yes, the msgid's are preserved.

po/perform_translations/translator.py
```diff
@@ -0,0 +1,106 @@
...
+for i in theActualArray:
+  entry = polib.POEntry(
+      msgid=i[0],
+      msgstr=i[1]
+  )
+  po.append(entry)
+
+po.save('newfile.po')
+
+subprocess.run('msgmerge newfile.po ' +  language + '.po ' + '> newfile.po', shell=True)
```
gcampax, line 102: Did you test this line? It cannot possibly work, because the shell will open newfile.po for writing (and truncate it) before it spawns msgmerge. Also, I thought I mentioned to avoid running commands through the shell...

Krish-sysadmin: My bad. Fixed it. Updated the README

gcampax:
Alright, this looks ok, let's merge this!

Issue #668, opened 2021-06-22
Fails to compute a reply to "currently playing" question


gcampax:
```
$dialogue @org.thingpedia.dialogue.transaction.execute;
[artists] of @com.spotify(id="com.spotify-gcampax"^^tt:device_id("Spotify Account of gcampax")).get_currently_playing()
#[results=[
  { id="spotify:track:67lbzVADsqJJ6tyVk0XCgn"^^com.spotify:song("Chain My Heart"), artists=["spotify:artist:0u6GtibW46tFX7koQ6uNJZ"^^com.spotify:artist("Topic"), "spotify:artist:64M6ah0SkkRsnPGtGiRAbb"^^com.spotify:artist("Bebe Rexha")] }
]];
Genie executed new program: uuid-28303eb8-9222-4899-9019-bf5fe73cba9e
--- DEPTH 0
stats: size(charts[0][constant_Number]) = 13
stats: size(charts[0][constant_String]) = 1
stats: size(charts[0][out_param_Any]) = 15
stats: size(charts[0][constant_Entity__com__spotify__artist]) = 2
stats: size(charts[0][constant_Entity__com__spotify__song]) = 1
stats: size(charts[0][constant_Enum__on__off]) = 2
stats: size(charts[0][constant_Enum__track__context__off]) = 3
stats: size(charts[0][ctx_with_result]) = 1
stats: size(charts[0][ctx_with_table_result]) = 1
stats: size(charts[0][ctx_without_action]) = 1
stats: size(charts[0][ctx_with_result_noquestion]) = 1
stats: size(charts[0][ctx_current_statement]) = 1
stats: size(charts[0][ctx_display_nonlist_result]) = 1
stats: size(charts[0][ctx_current_query]) = 1
stats: size(charts[0][ctx_thingpedia_result]) = 1
stats: size(charts[0][ctx_result_name]) = 1
depth 0 took 14 milliseconds

--- DEPTH 1
stats: size(charts[1][constant_Any]) = 22
stats: size(charts[1][actionable_system_display_result_short]) = 1
depth 1 took 1 milliseconds

--- DEPTH 2
stats: size(charts[2][one_param_direct_result_info_phrase]) = 1
depth 2 took 1 milliseconds

--- DEPTH 3
stats: size(charts[3][direct_result_info_phrase]) = 1
depth 3 took 2 milliseconds

--- DEPTH 4
depth 4 took 2 milliseconds

--- DEPTH 5
[...]

PR #748, opened 2021-08-09
Add templates for direct result phrases with array parameters

gcampax:
"Direct result phrases" are phrases of the form "The X is Y."

Add the variants of these phrases for arrays, so we can query
the artists of the currently playing song. We need this template
because we don't use the form "I found X. It is a song written by Y" if
the query is not a list.

Addresses #668


gcampax:
@rayslxu it looks like this PR has been sitting since I went on vacation, can you take a look when you have time?

rayslxu (approved):

PR #751, opened 2021-08-13
Save and restore conversation state in db

emilyjchang:


lib/dialogue-agent/assistant_dispatcher.ts
```diff
@@ -161,11 +164,22 @@ export default class AssistantDispatcher extends events.EventEmitter {
...
+        const state = await this._conversationStateDB.getOne(conversationId).then((row) => {
+            if (row) {
+                return {
+                    history : row.history ? JSON.parse(row.history) : [],
+                    dialogueState : row.dialogueState ? JSON.parse(row.dialogueState) : null,
+                    lastMessageId : row.lastMessageId ? row.lastMessageId : 0
+                };
+            }
+            return undefined;
+        });
```
gcampax, line 176: This is inside converse() which is used by the REST endpoint /me/api/converse. We'll need equivalent code in getOrOpenConversation (or a new API) so the usual WebSocket endpoint also uses it.

lib/dialogue-agent/conversation.ts
```diff
@@ -79,6 +80,7 @@ export interface ConversationDelegate {
 export interface ConversationState {
     history : Message[];
     dialogueState : Record<string, unknown>;
+    lastMessageId : number|null;
```
gcampax, line 83: Can this be ever null? Why? Let's make 0 or 1 the default.

lib/dialogue-agent/conversation.ts
```diff
@@ -305,6 +310,18 @@ export default class Conversation extends events.EventEmitter {
         if (this._history.length > 30)
             this._history.shift();
         await this._callDelegates((out) => out.addMessage(msg));
+
+        this.saveState(msg.id);
+    }
+
+    saveState(lastMessageId : number) {
```
gcampax, line 317: Async here, and await the insertOne call to handle errors correctly.

lib/dialogue-agent/conversation.ts
```diff
@@ -305,6 +310,18 @@ export default class Conversation extends events.EventEmitter {
         if (this._history.length > 30)
             this._history.shift();
         await this._callDelegates((out) => out.addMessage(msg));
+
+        this.saveState(msg.id);
```
gcampax, line 314: Await this call or catch the error and log it.

lib/engine/db/sqlite/migration.ts
```diff
@@ -99,6 +99,15 @@ function migrateTo9(db : sqlite3.Database) {
                'vote text default null, ' +
                'comment text default null)');
     });
+},
+function migrateTo10(db : sqlite3.Database) {
+    db.serialize(() => {
+        db.run('dop table if exists conversation_state');
```
gcampax, line 105: Typo: "dop"

gcampax (changes requested):
Thanks for working on this!

We should add some tests in test/engine probably.
Tests in test/agent won't be sufficient because the engine/db stuff are mocked out.

lib/dialogue-agent/conversation.ts
emilyjchang: In that case should the default db value be 0 as well?

gcampax: It doesn't really matter, we'll always write a value to the row, but yeah it could be. (I don't remember if sqlite supports default clauses though)

PR #752, opened 2021-08-26
Update dialogue handler confidence flow

jmhw0123:


gcampax (approved):
I think the logic is correct now!

PR #755, opened 2021-08-26
Add join templates to genie

rayslxu:
Mainly focus on the self joins as they are considered more common.

This PR requires stanford-oval/thingtalk#380


lib/templates/dialogue_acts/coref-questions.ts
```diff
@@ -227,6 +227,7 @@ function booleanQuestion(base : Ast.Expression|null, slot : C.FilterSlot) : C.Pa
             schema: slot.schema,
             type: slot.ptype,
             filterable: slot.schema.getArgument(op1.name)!.getImplementationAnnotation<boolean>('filterable') ?? true,
+            symmetric: slot.schema.getArgument(op1.name)!.getImplementationAnnotation<boolean>('symmetric') ?? false,
```
gcampax, line 230: What does symmetric mean?

lib/templates/joins.genie
```diff
@@ -0,0 +1,86 @@
...
+    | table:short_phrase_table 'pairs' ('such that'|'that') pronoun_the_first ('is'|'are'|'was'|'were') ('the'|'a'|'') param:out_param_Any 'of' pronoun_the_second 
+    | table:short_phrase_table 'pairs' ('such that'|'that') pronoun_the_first ('has'|'have'|'had') pronoun_the_second 'as' ('the'|'a'|'') param:out_param_Any 
+    ) => C.makeSelfJoinFromParam($loader, table, param);
+
+    ( 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first ('has'|'have'|'had') condition:npp_join_condition
+    | 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first ('is'|'are'|'was'|'were') ('the'|'a'|'') condition:npi_join_condition
+    | 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first ('is'|'are'|'was'|'were') condition:pvp_join_condition 
+    | 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first ('is'|'are'|'was'|'were') condition:preposition_join_condition 
+    | 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first condition:avp_join_condition
```
gcampax, line 57: typos: paris

lib/templates/joins.genie
```diff
@@ -0,0 +1,86 @@
...
+}
+
+generic_join_table = { 
+    table1:short_phrase_table ('and'|'with') 'the' table2:base_table_hidden param2:out_param_Any 'of' ('its'|'their') param1:out_param_Any
+    => C.makeGenericJoin($loader, table1, param1, table2, param2);
+}
+
+complete_table : Ast.Expression = {
+    self_join_table;
+    generic_join_table;
```
gcampax, line 85: Please gate these under "notablejoin" so I can disable them in production and multiwoz

lib/templates/joins.genie
```diff
@@ -0,0 +1,86 @@
...
+    complete_table;
+    one_clean_filter_table;
+}
+
+self_join_table : Ast.Expression = {
+    ( 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first ('is'|'are'|'was'|'were') ('the'|'a'|'') param:out_param_Any 'of' pronoun_the_second 
+    | 'paris of' table:short_phrase_table ('such that'|'that') pronoun_the_first ('has'|'have'|'had') pronoun_the_second 'as' ('the'|'a'|'') param:out_param_Any 
+    | table:short_phrase_table 'pairs' ('such that'|'that') pronoun_the_first ('is'|'are'|'was'|'were') ('the'|'a'|'') param:out_param_Any 'of' pronoun_the_second 
+    | table:short_phrase_table 'pairs' ('such that'|'that') pronoun_the_first ('has'|'have'|'had') pronoun_the_second 'as' ('the'|'a'|'') param:out_param_Any 
+    ) => C.makeSelfJoinFromParam($loader, table, param);
```
gcampax, line 51: So this generates (among other):
"male people pairs such that the first is the spouse of the second"
is that the expected sentence?
"X pairs" sounds awkward to me, especially with an adjective in front.

It also generates
"male people pairs that the first is the spouse of the second"
(using "that" instead of "such that").
I think this one is just wrong.

It also generates
"people born in X pairs such that the first is the spouse of the second"
(using the "pvp" form of one_clean_filter_table)
I think this one is also just wrong.

gcampax (commented):
Looks good, a couple of comments.

lib/templates/dialogue_acts/coref-questions.ts
rayslxu: for a property `p` in table `t`, `p` is considered "symmetric" if `p` has the same entity type the id of `t`, and if `x` is in `t filter p == y`, then `y` is in `t filter p == x`.

rayslxu: also applies to array types, where the second condition would be if `x` is in `t filter contains(p, y)`, then `y` is in `t filter contains(p, x)`.

rayslxu: examples: `sister_city`, `siblings`

lib/templates/joins.genie
rayslxu: yeah you are right I removed the whole "X pairs" thing from the templates

lib/templates/keyfns.ts
```diff
@@ -74,14 +74,16 @@ export function inputParamKeyFn(slot : InputParamSlot) {
...
     return {
         functionName: slot.schema.qualifiedName,
         type: slot.type,
         is_numeric: slot.type.isNumeric(),
         elem: slot.type instanceof Type.Array ? slot.type.elem as Type : null,
         is_numeric_elem: slot.type instanceof Type.Array ? (slot.type.elem as Type).isNumeric() : false,
         associatedIdType: id && !id.is_input ? id.type : null,
-        filterable: slot.filterable
+        filterable: slot.filterable,
+        symmetric: symmetric ? (symmetric as Ast.BooleanValue).value : false
```
gcampax, line 86: This doesn't look right. getImplementationAnnotation calls toJS on the annotation, so symmetric will be a boolean or undefined.

lib/templates/keyfns.ts
```diff
@@ -74,14 +74,16 @@ export function inputParamKeyFn(slot : InputParamSlot) {
 
 export function paramKeyFn(slot : ParamSlot) {
     const id = slot.schema.getArgument('id');
+    const symmetric = slot.schema.getArgument(slot.name)?.getImplementationAnnotation('symmetric');
```
gcampax, line 77: If you can make this typesafe (ish) if you do `getImplementationAnnotation<boolean>('symmetric')`

gcampax (approved):
Looks good minus one thing to check

PR #777, opened 2021-09-11
Fix issue with functions that need to call other functions

Anggriosutopo:
Fix the issue for generating training data for cases where the function needs to call another function.


gcampax (approved):
Yep, looks good, thank you!

PR #783, opened 2021-09-12
Part 1 of fixing timezone issues

gcampax:
This is the first PR that attempts to fix all issues related to timezone, which cause us to do timers at the wrong time when the system timezone does not match the user timezone (always the case in cloud).

The PR makes use of the Temporal stage 3 EcmaScript proposal. It will soon be part of the JS standard, and is used as a polyfill. I chose this library instead of relying on other third-party libraries like moment because when it becomes part of the standard we can get rid of the dependency and rely on v8/node doing the right thing right away instead.


gcampax:
I tested this with almond-server with various settings of TZ variable and some hacks to change the timezone. It fixes the issues with alarm, and the thingtalk code becomes consistent with the sentence. I just realized, in many case the code was not consistent with the agent utterance because the time in the code would be UTC or some random timezone, and the agent utterance would be in the user's timezone. Now everything is in the user's configured timezone.

I still need to debug the problem with the scenario tests.

rayslxu (approved):
looks good

PR #787, opened 2021-09-16
Add a new audio control protocol

gcampax:
This protocol allows the server to control playback on speaker clients connected over websocket. It will be used to let almond-server & almond-cloud control the audio of genie-client-cpp.

It supports:
- waiting until audio is ready to play (so replies don't overlap with new music)
- volume control
- playing URLs (for Tunein)
- single word "pause" and "resume" that affect the last playing skill (not just spotify/skills that implement the media player interface)

Includes some general cleanups of the whole server side of the websocket protocol, allowing more code reuse between almond-server and almond-cloud.


nrser (approved):
seems reasonable

PR #790, opened 2021-09-20
Fix dbquery

rayslxu:


gcampax (approved):
Yeah definitely. We should add the signature of `query` to the declaration of BaseDevice in thingpedia-api:
```typescript
interface BaseDevice {
  query?(query : ThingTalk.Ast.Program, env : ThingTalk.Runtime.ExecEnvironment) : QueryFunctionResult;
}
```
where QueryFunctionResult is defined as in this file:
```typescript
type QueryResult = Record<string, unknown>;
type QueryFunctionResult = Iterable<QueryResult>|AsyncIterable<QueryResult>|Promise<Iterable<QueryResult>>;
```
(in fact, the QueryFunctionResult is not defined exactly this way in this file but it's not quite correct, it's missing the non AsyncIterable cases)

PR #793, opened 2021-09-29
Starter code for MTOD

Mehrad0711:
Bare minimum code to generate a single-domain dataset for MTOD (successor of BiTOD) mimicking multiwoz. 
I'm also figuring a script to convert bitod annotations similar to "auto-annotate-multiwoz". Once I become more familiar with the dialogue code I can tweak the code/ templates to generate better examples.


gcampax:
Can we put this in a different repository? I'd rather not overload genie-toolkit with too many starters - tests are already too long and cloning the repo would be a pain.

Mehrad0711:
Yeah let's deal with this later.

PR #805, opened 2021-10-05
Dialogue PL, part 1

gcampax:
All the way to running the agent and synthesizing dialogues, without regressions, and hooking up the existing transaction model, but while keeping most of the logic in the form of templates.


gcampax:
The PR is not finished because synthesis is not done, but I got to the point where the agent works at inference time, and the agent tests pass, so all functionality is there. I think it's a good milestone to start the review.
@rayslxu do you want to take a look at this? Maybe we can schedule some time to go over it in person.

gcampax:
Alright, synthesis is not done yet but at least tests pass now. @rayslxu how do you want to review this? Do we want to merge this first PR somewhere?

rayslxu:
> Alright, synthesis is not done yet but at least tests pass now. @rayslxu how do you want to review this? Do we want to merge this first PR somewhere?

Ugh... this is a lot to read. I guess I will just skim through it to get a sense of what's going on... (unless you will split the PR into smaller ones)

PR #806, opened 2021-10-05
Wip/translation api

Mehrad0711:


Mehrad0711:
>Do we even need this PR, given we're not doing translation at test time?

I think having a translation endpoint wouldn't hurt even if we don't use it immediately.
For po translations, I'm planning to move the genienlp calls from makefile to JS code for which the `translateUtterance` method is useful.  
Also we spent so much time on it, we might as well merge it 🤣

gcampax:
The problem is that once you merge an API, you have to support it forever (until the next major API break). You have to compare the small amount of work you did so far, with the future amount of work.

PR #811, opened 2021-10-07
Enforce a consistent code style

gcampax:
@Mehrad0711 @rayslxu: @nrser suggested we use some automated tool to format our source code.
Some time ago I spent some time to configure eslint to be more or less what we use as code style, so this PR enforces that.

You can set up eslint as a formatter in VSCode (in the extension settings: Eslint > Format: Enable), and then autoformat with Ctrl+Shift+I
Alternatively you can fix all formatting issues with `npm run lint -- --fix` before committing.

Opinions? Suggestions?


rayslxu:
love that!

Mehrad0711:
I think it would be better if we can set it up as pre-commit-hook so we don't have to run it before every commit. But we can add that later.

nrser:
I'm obviously a fan... I think that's pretty much what I have setup on stanford-oval/almond-cloud with the vscode config:

```json
{
  "eslint.validate": [
    "javascript",
    "javascriptreact",
    "typescript",
    "typescriptreact",
  ],
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true,
  },
}
```

PR #814, opened 2021-10-12
requote: check end of spans too

Mehrad0711:
Just when I thought we addressed all corner cases...


Mehrad0711:
Updated the test example

PR #846, opened 2021-12-02
Update augmentation code

gcampax:
Making use of the new samplers I just added in https://github.com/stanford-oval/thingpedia-api/pull/213

Tagging Mehrad cause he touched augmentation last.


Mehrad0711 (approved):
OK

PR #872, opened 2022-01-31
Fix arguments after adding `-url`

Mehrad0711:
Fix bug introduced in https://github.com/stanford-oval/genie-toolkit/commit/711dfb96078ed2df4a5c62b67660454a0fe0976a


Mehrad0711:
@rayslxu same test is failing on master too. Should we fix that first?

rayslxu:
> @rayslxu same test is failing on master too. Should we fix that first?

It's fixed now.

PR #879, opened 2022-02-22
Remove almost all dialogue templates

gcampax:
As requested by @monicalam 

This PR removes all non-essential dialogue templates. Remaining are the basic yes/no/cancel and slot filling answers, confirmation and slot filling questions, and result phrases of various types.


rayslxu:
I don't think we want this to be merged into master any time soon

gcampax:
You should talk to Monica about this. It was her idea to remove everything.

PR #884, opened 2022-03-24
add support for measurement type

jmhw0123:
Added support for the measurement type


rayslxu (commented):
don't we need to update sample-constants first?

rayslxu (changes requested):
As it gets more and more complicated, I think we should have a unit test for it. Use a small fake schema with different thingtalk types to test it. Probably a good idea to have a parameter to control how many examples per annotation - so we can set it to 1 to make the unit test shorter.

rayslxu (approved):
looks good

